PR #148, opened 2023-10-23
Post review changes to #138

jamienoss:
See #138

### DIFF
--- a/biospecdb/apps/uploader/models.py
+++ b/biospecdb/apps/uploader/models.py
@@ -196,6 +196,14 @@ class Visit(DatedModel):
         """ Model validation. """
         super().clean()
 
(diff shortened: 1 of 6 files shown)


vkluzner (approved):
Looks fine to me. Liked that you created separate function for it. The only question is: why do we need "AUTO_FIND_PREVIOUS_VISIT = True"? Isn't it essential to always find the last visit since we have this field defined in `Visit` model?

jamienoss:
>The only question is: why do we need AUTO_FIND_PREVIOUS_VISIT = True? Isn't it essential to always find the last visit since we have this field defined in Visit model?

Well, I think we're going to nuke ``Visit.previous_visit`` but also, it's not a good idea for us to introduce behavior that can't necessarily guarantee correctness. With a config setting, we can add the code and if we find it doesn't work or the user doesn't want it, we/they can just turn it off.

PR #150, opened 2023-10-26
Update docker build

jamienoss:
## Changes:
 * SQLite db files are now created in ``db/`` rather than the project root. This is so that we can mount a persistent docker volume and create them there instead.
 * Use docker-compose.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
 FROM python:3.11-slim
 
+ENV PYTHONUNBUFFERED=1
+
 WORKDIR /app
 
-COPY requirements/prd.txt requirements.txt
+COPY requirements/prd.txt /app/requirements.txt
 
 RUN pip3 install -r requirements.txt
 
-COPY . .
-
-CMD ["uvicorn", "biospecdb.asgi:application", "--host", "0.0.0.0", "--port", "8000"]
+COPY manage.py /app/
+COPY biospecdb/ /app/biospecdb/
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,3 +10,5 @@ RUN pip3 install -r requirements.txt
 
 COPY manage.py /app/
 COPY biospecdb/ /app/biospecdb/
+
+CMD ["python", "manage.py", "runserver", "0.0.0.0:8000"]
--- a/biospecdb/apps/uploader/views.py
+++ b/biospecdb/apps/uploader/views.py
@@ -7,8 +7,9 @@ from .forms import FileUploadForm, DataInputForm
 from uploader.models import Patient, Visit, SpectralData, BioSample, Symptom, Disease
 from biospecdb.util import is_valid_uuid, to_uuid
 
+
 def home(request):
-    return render(request, 'home.html')
+    return render(request, 'Home.html')
 
 
 @staff_member_required
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,48 @@
+version: '3.8'
+
+services:
+  db-setup:
+    image: bsr
+    build:
+      context: .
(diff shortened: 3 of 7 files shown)


biospecdb/apps/uploader/views.py
```diff
@@ -7,8 +7,9 @@
 from uploader.models import Patient, Visit, SpectralData, BioSample, Symptom, Disease
 from biospecdb.util import is_valid_uuid, to_uuid
 
+
 def home(request):
-    return render(request, 'home.html')
+    return render(request, 'Home.html')
```
jamienoss, line 12: This fixes an existing bug as the linux fs is case-sensitive the templates could not be found. We really should add some client tests...

docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+        condition: service_completed_successfully
+    image: bsr
+    restart: unless-stopped
+    ports:
+      - "8000:8000"
+    volumes:
+      - db-data:/app/db
+    environment:
+      DJANGO_SETTINGS_MODULE: biospecdb.settings
+    command: python manage.py runserver 0.0.0.0:8000
```
jamienoss, line 41: A few things aren't rendering correctly when using uvicorn so reverting to just using django's server for now and will fix in #151.

scripts/dev/rebuild_db.sh
```diff
@@ -1,20 +1,31 @@
...
+
+# Delete uploaded data files.
 rm biospecdb/spectra_data/*
 rm biospecdb/raw_data/*
-rm admin.sqlite3 bsr.sqlite3
+
+# Delete databases.
+rm *.sqlite3
+rm db/*.sqlite3
+mkdir db
```
amitschang, line 12: perhaps better to reverse order and add `-p` to mkdir

```sh
mkdir -p db
rm db/*.sqlite3
```

Dockerfile
```diff
@@ -1,11 +1,12 @@
...
 WORKDIR /app
 
-COPY requirements/prd.txt requirements.txt
+COPY requirements/prd.txt /app/requirements.txt
 
 RUN pip3 install -r requirements.txt
 
-COPY . .
-
-CMD ["uvicorn", "biospecdb.asgi:application", "--host", "0.0.0.0", "--port", "8000"]
```
amitschang, line 11: maybe keep the command spec in there if at least for documentation and simplicity if just firing up lone container?

amitschang (commented):
Looks good to me, just a few comments

scripts/dev/rebuild_db.sh
```diff
@@ -1,20 +1,31 @@
...
 rm biospecdb/apps/user/migrations/0*
 rm biospecdb/apps/uploader/migrations/0*
+
+# Delete uploaded data files.
 rm biospecdb/spectra_data/*
 rm biospecdb/raw_data/*
-rm admin.sqlite3 bsr.sqlite3
+
+# Delete databases.
+rm *.sqlite3
```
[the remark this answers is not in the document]
jamienoss: True it could do, but perhaps as equally as the user still having the old db files lying around, so I added them both. We can remove down the road.
Opened #152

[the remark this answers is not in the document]
jamienoss: ...and added an inline comment ref'ing #152 .

Dockerfile
jamienoss: Will do, however, I'll add the django server invocation for now instead.

jamienoss: Well actually, now I remember why I removed it - it won't work as the DB hasn't been created yet.

@amitschang  Still think it should go in?

amitschang (approved):

PR #156, opened 2023-10-30
Close file before attempting to remove from fs (required for windows)

jamienoss:
This may be replaced by #158 

[os.remove](https://docs.python.org/3/library/os.html#os.remove):
>On Windows, attempting to remove a file that is in use causes an exception to be raised; on Unix, the directory entry is removed but the storage allocated to the file is not made available until the original file is no longer in use.

We must therefore ensure that files are closed before nuking.

Really this should be adequately context managed, however, it appears not to be. I had wondered if ``ContentFile`` was the culprit but its ``close`` is a NoOp ([src](https://github.com/django/django/blob/main/django/core/files/base.py#L141)).


glemson (dismissed):

glemson (approved):
Code changes seem sensible and I tested bulk upload which succeeded.

PR #167, opened 2023-11-03
Store spectral data file on file as json and in memory as dataclass

jamienoss:
# Changes:

### Principle:
 * Adds ``patient_id`` to spectral data file saved to FS storage.
 * Create ``SpectralData`` dataclass as our internal API to store spectral data in memory.
 * Clean all uploaded spectral data files by converting to a ``SpectralData`` dataclass instance and then saving to FS storage as json.

### Required refactoring:
 * Expand all file reading to accommodate json files, not just csv and xlsx (excel).
 * Explicit namespace import of ``uploader.io`` instead of ``from uploader.io import ...`` prevents namespace clashes and increases clarity that the code is ours.
 * Add unique ID as pk to ``uploader.models.SpectralData`` to facilitate unique filenames without having to commit to DB. Note: Without this, unique PKs are prescribed only by the DB and not the django layer thus requiring a save - resulting in a chicken-and-egg situation when cleaning files and needing their final filenames since cleaning happens before and independently before saving.
 * Refactor bulk upload, prescribing unique filename to data files.
 * Refactor ``io`` funcs to help unify API by accepting only a ``file`` param instead of ``file, ext``. This moves all calls to ``uploader.io.get_file_info`` internal to io funcs rather than being called externally. ``get_file_info`` could be made protected.


biospecdb/apps/uploader/tests/test_models.py
```diff
@@ -293,8 +293,8 @@ def test_temp_files_deleted(self, mock_data_from_files):
         filename = Path(SpectralData.objects.all()[0].data.name)
         assert filename.parent.exists()
         assert filename.parent.is_dir()
-        assert len(list(filename.parent.glob('*'))) == n_patients
         assert not list(filename.parent.glob(f"*{TEMP_FILENAME_PREFIX}*"))
+        assert len(list(filename.parent.glob('*'))) == n_patients
```
jamienoss, line 297: Better ordering. If temp files exist the glob length will most likely never equal ``n_patients``. Similarly, if its length never equals ``n_patients`` there's a good chance that it's because temp files still exist (or file duplication).

biospecdb/apps/uploader/migrations/0001_initial.py
```diff
@@ -164,11 +164,12 @@ class Migration(migrations.Migration):
...
-                ('spectral_data_file', models.FileField(help_text='File containing rows of spectral intensities for the corresponding meta data file.', upload_to='raw_data/', validators=[django.core.validators.FileExtensionValidator(['csv', 'xlsx'])])),
+                ('meta_data_file', models.FileField(help_text='File containing rows of all patient, symptom, and other meta data.', upload_to='raw_data/', validators=[django.core.validators.FileExtensionValidator(['csv', 'xlsx', 'json'])])),
+                ('spectral_data_file', models.FileField(help_text='File containing rows of spectral intensities for the corresponding meta data file.', upload_to='raw_data/', validators=[django.core.validators.FileExtensionValidator(['csv', 'xlsx', 'json'])])),
                 ('center', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, to='uploader.center')),
             ],
             options={
+                'verbose_name': 'Bulk Data Upload',
```
jamienoss, line 172: A missed vestige of a prior PR.

biospecdb/apps/uploader/tests/test_exporters.py
```diff
@@ -72,14 +73,17 @@ def test_with_data_files_check_content(self, csv_export):
...
             assert set(data[SpectralData.data.field.name]) == set(spectral_data_files)
 
         for filename in spectral_data_files:
-            data = spectral_data_from_csv(filename)
-            assert set(data.columns) == {"wavelength", "intensity"}
-            assert len(data) == 1798
+            data = uploader.io.spectral_data_from_json(filename)
+            assert isinstance(data, uploader.io.SpectralData)
+
+            assert {x.name for x in dataclasses.fields(data)} == {"patient_id", "wavelength", "intensity"}
```
amitschang, line 84: maybe it makes more sense in this test to just assert `data.patient_id == {whatever}`?

biospecdb/apps/uploader/tests/test_io.py
```diff
@@ -0,0 +1,93 @@
...
+
+@pytest.fixture(scope="module")
+def json_data():
+    filename = SPECTRAL_FILE_PATH.with_suffix(uploader.io.FileFormats.JSON)
+    with open(filename, mode="r") as fp:
+        return json.load(fp)
+
+
+class TestReadSingleRowSpectralDataTable:
+    @pytest.mark.parametrize("ext", (uploader.io.FileFormats.CSV, uploader.io.FileFormats.XLSX))
```
amitschang, line 22: Is JSON meant to be omitted? Maybe I read wrong but looked like the read function handled all these formats

biospecdb/apps/uploader/tests/test_io.py
```diff
@@ -0,0 +1,93 @@
...
+class TestReadRawData:
+    def test_read_raw_data_ext_exception(self):
+        with pytest.raises(ValueError, match="When passing an IO stream, ext must be specified"):
+            uploader.io.read_raw_data(None, None)
+
+        with pytest.raises(ValueError, match="A path-like or file-like object must be specified"):
+            uploader.io.read_raw_data(None, ext=".whatever")
+
+        with pytest.raises(NotImplementedError, match="File ext must be one of"):
+            uploader.io.read_raw_data("somefile", ext=".whatever")
```
amitschang, line 93: its a small thing but perhaps assert `'.whatever'` not in `uploader.io.FileFormats.list()` is good protection

biospecdb/apps/uploader/tests/test_models.py
```diff
@@ -288,19 +289,46 @@ def test_files_added(self, mock_data_from_files):
             assert Path(obj.data.name).exists()
 
     def test_temp_files_deleted(self, mock_data_from_files):
+        n_patients = 10
+        assert len(SpectralData.objects.all()) == n_patients
+        filename = Path(SpectralData.objects.all()[0].data.name)
+        assert filename.parent.exists()
+        assert filename.parent.is_dir()
+        assert not list(filename.parent.glob(f"*{uploader.io.TEMP_FILENAME_PREFIX}*"))
```
amitschang, line 297: should there not be a `*` at the front if its a prefix?

amitschang (approved):
Some comments inline, some of which I think are worth addressing

amitschang (commented):
I meant to select comment

biospecdb/apps/uploader/charts.py
```diff
@@ -57,11 +57,12 @@ def get_line_chart(result: "QueryResult") -> Optional[str]:  # noqa: F821
...
                           yaxis_title="Intensity",
                           title=f"Spectral Data for SQL query: '{result.sql}'")
-        for _, (patient_id, filename) in df.iterrows():
-            spectral_data = spectral_data_from_csv(filename)
-            fig.add_scatter(x=spectral_data["wavelength"],
-                            y=spectral_data["intensity"],
-                            name=patient_id)
+        for row in df.itertuples():
+            spectral_data = uploader.io.read_spectral_data(row.data)
+            assert spectral_data.patient_id != row.patient_id  # Internal sanity check.
```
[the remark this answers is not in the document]
jamienoss: Good catch! I originally had this as an ``if`` and then changed it without inverting the logic. The tests never failed because, whilst the ID strings were the same, one was a ``UUID`` instance and the other just a string whose comparison fails.

biospecdb/apps/uploader/io.py
```diff
@@ -50,17 +76,24 @@ def read_raw_data(file, ext=None):
         data = pd.read_csv(file, **kwargs)
     elif ext == FileFormats.XLSX:
         data = pd.read_excel(file, **kwargs)
+    elif ext == FileFormats.JSON:
+        data = pd.read_json(file, dtype=kwargs["dtype"])
+        data.replace(kwargs["true_values"], True, inplace=True)
+        data.replace(kwargs["false_values"], False, inplace=True)
+        na_values = kwargs["na_values"].copy()
+        na_values.extend(list(STR_NA_VALUES))
+        data.replace(na_values, None, inplace=True)
```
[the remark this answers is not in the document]
jamienoss: >but keeping the None would force an object type in dataframe

Good catch again.

I think ``pd.NA`` should be sufficient without the need to convert types since we only want to replicate the parsing taking place in ``pd.read_csv`` and  ``pd.read_excel`` which both take ``na_values`` args. I think...

>any potential need to preserve actual numeric NaN

Perhaps, but at this stage, I don't think so. Either way, that would/could be beyond the scope of this PR as it also affects the other two readers (``pd.read_csv`` and  ``pd.read_excel``) that this PR tries not to touch.

biospecdb/apps/uploader/io.py
```diff
@@ -77,38 +110,135 @@ def read_meta_data(file, ext=None):
...
 
-def read_spectral_data_table(file, ext=None):
-    data = read_raw_data(file, ext=ext)
+def read_spectral_data_table(file):
+    """ Read in multiple rows of data returning a pandas.DataFrame.
+
+        This assumes the following format:
+        | patient_id | min_lambda | ... | max_lambda |
+        | ---------- | ---------- | --- | ---------- |
+        |<some UUID> | intensity  | ... | intensity  |
```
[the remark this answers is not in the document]
jamienoss: Is this better/sufficient?
>The data to be read in, needs to be of the following table layout:
Note: Commas need to be present for CSV data.
Note: The following docstring uses markdown table syntax.

biospecdb/apps/uploader/io.py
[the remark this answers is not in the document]
amitschang: `convert_datatypes` is probably still good, since during the reading phase it will have to cast to str to handle the null value keywords. The replacement wont put that back so would remain a string (in fact, perhaps the pd.NA vs None is not as important since I think the conversion will turn None into pd.NA)

biospecdb/apps/uploader/tests/test_exporters.py
jamienoss: @amitschang Is this comment on the wrong line as the two tests aren't equivalent? As is, we're testing that the fields are as expected, not their values.

amitschang: I meant that we can both test that `patient_id` is in the object and that it has a reasonable value at the same time using an assert on its value. Doesn't seem as important to me to ensure the data structure has strictly those fields.

biospecdb/apps/uploader/io.py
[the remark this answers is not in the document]
amitschang: Yeah, but perhaps edit like

```
...following table layout (e.g. each row refers to a patient with a patient_id column 
followed by a column for each wavelength channel)...
```

It would probably also be good to be clear about if headers are expected, etc.

And maybe add to 

```
Note: The following docstring uses markdown table syntax (i.e. it is not the literal format 
expected, which should be one of ...).
```

or so.

biospecdb/apps/uploader/tests/test_io.py
amitschang: True, but that does make an assumption on the behavior of the function, which is what the test is testing. It seems really unlikely that whatever ends up there in any case.

biospecdb/apps/uploader/io.py
```diff
@@ -41,6 +43,28 @@ def choices(cls):
...
+    # Note: Some (if not all) pandas funcs return None when `inplace=True` instead of a ref to the passed df, making for
+    # awkward assignment due to the conditional return.
+
+    cleaned_df = df
+
+    _cleaned_df = cleaned_df.rename(columns=lambda x: x.lower() if isinstance(x, str) else x, inplace=inplace)
+    cleaned_df = cleaned_df if inplace else _cleaned_df
+
+    _cleaned_df = cleaned_df.rename(columns={"patient id": PATIENT_ID_STR}, inplace=True, errors="ignore")
+    cleaned_df = cleaned_df if inplace else _cleaned_df
```
ryanhausen, line 56: I don't think we need the tertiary statement here if `inplace` is always `True`

biospecdb/apps/uploader/io.py
```diff
@@ -41,6 +43,28 @@ def choices(cls):
...
+    # Note: This func is used to read in the data table returned from an SQL query from the explorer app and therefore,
+    # may not contain a PATIENT_ID_STR column - ``dropna`` has no ``errors`` arg.
+    if PATIENT_ID_STR in cleaned_df.columns:
+        # TODO: Raise on null patient id instead of silently dropping possible data.
+        _cleaned_df = cleaned_df.dropna(subset=[PATIENT_ID_STR], inplace=inplace)
+        cleaned_df = cleaned_df if inplace else _cleaned_df
+
+    return cleaned_df
+
+
```
ryanhausen, line 67: I don't have a high level view of the project yet, so feel free to discard this.

If this function only accepts dataframes and offers the same `inplace` argument with the same behavior that pandas offers, should we follow the same expected return value and only return something if `inplace==False`?

biospecdb/apps/uploader/io.py
jamienoss: Good catch, it shouldn't be always ``True`` but ``inplace=inplace``. Thanks!

biospecdb/apps/uploader/io.py
jamienoss: >should we follow the same expected return value and only return something if inplace==False

I'd very much prefer not to have any conditional returns as they make code cumbersome.

biospecdb/apps/uploader/io.py
ryanhausen: ok sound good, I think my last thought on this might be that renaming the `"patient id"` to `PATIENT_ID_STR` only matters if column exists and if the column does exist, then the condition below is also satisfied. So you could move it into the scope of the condition and not have to worry about errors:

``` python
    # Note: This func is used to read in the data table returned from an SQL query from the explorer app and therefore,
    # may not contain a PATIENT_ID_STR column - ``dropna`` has no ``errors`` arg.
    if "patient id" in cleaned_df.columns:
        # TODO: Raise on null patient id instead of silently dropping possible data.
        _cleaned_df = cleaned_df.rename(columns={"patient id": PATIENT_ID_STR}, inplace=True)
        cleaned_df = cleaned_df if inplace else _cleaned_df
        
        _cleaned_df = cleaned_df.dropna(subset=[PATIENT_ID_STR], inplace=inplace)
        cleaned_df = cleaned_df if inplace else _cleaned_df
```

amitschang: Is there a particular reason to support using `inplace=True`?

jamienoss: Perf

amitschang: by how much?

jamienoss: 🤷 We know upfront that we don't need to copy so...

amitschang: I don't know. In that case then just always `inplace=True`? But then I think the operation chaining without it can be more readable

jamienoss: >But then I think the operation chaining without it can be more readable

Totally agree there, which is why conditional returns aren't cool as they require conditional return assignment.

jamienoss: Is it really that bad?

amitschang: Probably not, but if it can make cleaner code (in this case I think it can) then I think it is worth it to use only one-or-the-other (and I lean toward not using `inplace=True`). Anyhow, I don't mean to block anything here, so I have no more comments on this :smile:

biospecdb/apps/uploader/io.py
ryanhausen: Stylewise something feels a little off about having a tertiary statement after each dataframe statement, but is an awkward thing to handle

jamienoss: I totally agree, it looks completely ugly, and this is exactly why conditional returns aren't cool.

biospecdb/apps/uploader/io.py
```diff
@@ -77,38 +132,167 @@ def read_meta_data(file, ext=None):
...
 
 
-def read_spectral_data_table(file, ext=None):
-    data = read_raw_data(file, ext=ext)
+def read_spectral_data_table(file):
+    """ Read in multiple rows of data returning a pandas.DataFrame.
+
+        The data to be read in, needs to be of the following table layout for .csv & .xlsx:
+        Note: Commas need to be present for CSV data.
+        Note: The following docstring uses markdown table syntax.
```
amitschang, line 140: I think napoleon is used for sphinx here, so it would be preferred to do:
```
Note:
    Commas need to be present for CSV data.
```
i.e. a blank line after `Note:` and indented content. This will render correctly in html docs.

biospecdb/apps/uploader/tests/test_io.py
```diff
@@ -0,0 +1,98 @@
...
+@pytest.fixture(scope="module")
+def json_data():
+    filename = SPECTRAL_FILE_PATH.with_suffix(uploader.io.FileFormats.JSONL)
+    with open(filename, mode="r") as fp:
+        return json.load(fp)
+
+
+class TestReadSingleRowSpectralDataTable:
+    @pytest.mark.parametrize("ext", uploader.io.FileFormats.list())
+    def test_read(self, json_data, ext):
```
amitschang, line 23: nitpick, its not actually `json_data` but probably better `data`, since it is the data from multiple formats.

amitschang (approved):
Looks good, thanks!

PR #172, opened 2023-11-21
Add nested inline forms for patient data input

jamienoss:
This uses [django-nested-inline](https://github.com/OskarPersson/django-nested-inline). However, it appears that this package seems fairly buggy. Namey permissions issues and also the inability to save files in multi-nested inlines, i.e., our spectral data files.

Looking at the package issues and open PRs, these are known issues, some with fixes in open PRs, however, I'm not optimistic that this is a good package for us to use.


jamienoss:
superseded by #179

PR #173, opened 2023-11-22
Replace symptom -> observation & disease -> observable

jamienoss:
Precursor to #117


jamienoss:
I've reviewed this twice now, no need to put someone else through the anguish.

PR #174, opened 2023-11-22
Add models.Observable.category

jamienoss:
Moved from #117 

Resolves [Notion ticket](https://www.notion.so/Update-model-observables-dc1991a270994165991b68aa1ae7c716?pvs=4)

In #117, it was suggested that we use fieldsets to separate these out in the form, however, it may be better to use list filters instead as is implemented in this PR. Fieldsets may look nicer, however, ``list_filter`` would still be needed, in which case, ~it's not clear to me how well they work together - might look odd.~ Momentary confusion, these are completely orthogonal since fieldsets work on fields. What would be useful is for different observation categories to be sectioned like fieldsets for the inline forms. This may also allow them to collapse.

### DIFF
--- a/biospecdb/apps/uploader/migrations/0001_initial.py
+++ b/biospecdb/apps/uploader/migrations/0001_initial.py
@@ -114,6 +114,7 @@ class Migration(migrations.Migration):
                 ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('created_at', models.DateTimeField(auto_now_add=True)),
                 ('updated_at', models.DateTimeField(auto_now=True)),
+                ('category', models.CharField(choices=[('BLOODWORK', 'Bloodwork'), ('COMORBIDITY', 'Comorbidity'), ('DRUG', 'Drug'), ('PATIENT_INFO', 'Patient Info'), ('PATIENT_INFO_II', 'Patient Info Ii'), ('PATIENT_PREP', 'Patient Prep'), ('SYMPTOM', 'Symptom'), ('TEST', 'Test'), ('VITALS', 'Vitals')], max_length=128)),
                 ('name', models.CharField(max_length=128)),
                 ('description', models.CharField(max_length=256)),
                 ('alias', models.CharField(help_text='Alias column name for bulk data ingestion from .csv, etc.', max_length=128)),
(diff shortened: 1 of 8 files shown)


biospecdb/apps/uploader/migrations/0001_initial.py
```diff
@@ -34,7 +34,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('observable', models.CharField(db_column='observable', max_length=128)),
                 ('value_class', models.CharField(choices=[('BOOL', 'Bool'), ('STR', 'Str'), ('INT', 'Int'), ('FLOAT', 'Float')], default='BOOL', max_length=128)),
-                ('days_observed', models.IntegerField(blank=True, default=None, null=True, validators=[django.core.validators.MinValueValidator(0)], verbose_name='Days of Observations onset')),
+                ('days_observed', models.IntegerField(blank=True, default=None, null=True, validators=[django.core.validators.MinValueValidator(0)], verbose_name='Days of symptoms onset')),
```
jamienoss, line 37: Whoops, looks like a 4th scan of #173 was needed, as I forgot to rebuild the migration after reverting this in ``models.py``.

biospecdb/apps/uploader/migrations/0001_initial.py
```diff
@@ -222,7 +223,7 @@ class Migration(migrations.Migration):
                 ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('created_at', models.DateTimeField(auto_now_add=True)),
                 ('updated_at', models.DateTimeField(auto_now=True)),
-                ('days_observed', models.IntegerField(blank=True, default=None, null=True, validators=[django.core.validators.MinValueValidator(0)], verbose_name='Days of Observations onset')),
+                ('days_observed', models.IntegerField(blank=True, default=None, null=True, validators=[django.core.validators.MinValueValidator(0)], verbose_name='Days of symptoms onset')),
```
jamienoss, line 226: Whoops, looks like a 4th scan of #173 was needed, as I forgot to rebuild the migration after reverting this in ``models.py``.

glemson (approved):
Seems ok. Was able to run the code, add a DRUG observable and add corresponding observation.

PR #178, opened 2023-11-28
Cast None as None in uploader.base_models.Types.cast

jamienoss:
Fields that are allowed to be both blank and/or null require this code.


jamienoss:
My memory was nagging me that we'd discussed needing this before and I wondered why we didn't. I went digging but only found the following PR, but there was nothing in there about ``None``. So 🤷

jamienoss:
@amitschang Hey, would you mind taking a quick look at this, please, cheers.

vkluzner (approved):

PR #179, opened 2023-11-28
Add nested inline forms to data admin

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Add-nested-inline-forms-to-Data-Admin-88187cd188834949b5f1fc20f99d4144?pvs=4).

This uses the 3rd party package [django-nested-admin](https://github.com/theatlantic/django-nested-admin/).

This PR supersedes #172 

Dependant on #178 

Out of the box, Django gives you inline admin forms, however, they're limited to the model's fields and the fields of any fk. It doesn't follow fks beyond the 1st layer, i.e., patient can have an inline for visit but none of visit's fk inlines. Nested inlines is needed to achieve data collection from a single admin page/form (without having to write this ourselves).
nested-inlines

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -4,6 +4,7 @@ from django.contrib import admin
 from django.db.models import Q
 from django.db.utils import OperationalError
 import django.forms as forms
+from nested_admin import NestedStackedInline, NestedTabularInline, NestedModelAdmin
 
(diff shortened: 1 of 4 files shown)


biospecdb/apps/uploader/base_models.py
```diff
@@ -38,6 +38,9 @@ class Types(TextChoices):
     FLOAT = auto()
 
     def cast(self, value):
+        if value is None:
```
jamienoss, line 41: See #178

amitschang (approved):
easy peasy!

PR #180, opened 2023-11-28
Restrict access to home to staff users (all pages now require login)

jamienoss:
@glemson yea or nay?


jamienoss:
I think it's safer to just lock everything down - we can always re-explore this another day.

PR #181, opened 2023-11-28
Check against null self.data.name rather than self.data

jamienoss:
The following exception was raised instead of a validation error:
```python
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../../django/django/db/models/base.py:1477: in full_clean
    self.clean()
../models.py:538: in clean
    if (cleaned_file := self.clean_data_file()) is not None:
../models.py:523: in clean_data_file
    data = uploader.io.read_spectral_data(self.data)
../io.py:265: in read_spectral_data
    _fp, filename = _get_file_info(file)
../io.py:291: in _get_file_info
    fp = file.file
../../../../../django/django/db/models/fields/files.py:46: in _get_file
    self._require_file()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <FieldFile: None>

    def _require_file(self):
        if not self:
>           raise ValueError(
                "The '%s' attribute has no file associated with it." % self.field.name
            )
E           ValueError: The 'data' attribute has no file associated with it.

../../../../../django/django/db/models/fields/files.py:41: ValueError
```

### DIFF
--- a/biospecdb/apps/uploader/models.py
+++ b/biospecdb/apps/uploader/models.py
@@ -514,7 +514,9 @@ class SpectralData(DatedModel):
 
             The schema is equivalent to `json.dumps(dataclasses.asdict(uploader.io.SpectralData))``.
         """
-        if self.data is None:
+
+        # Note: self.data is a FieldFile and is never None so check is "empty" instead, i.e., self.data.name is None.
+        if not self.data:
             return
 
         try:
--- a/biospecdb/apps/uploader/tests/test_models.py
+++ b/biospecdb/apps/uploader/tests/test_models.py
@@ -296,6 +296,14 @@ class TestSpectralData:
         for obj in SpectralData.objects.all():
             assert Path(obj.data.name).exists()
 
+    def test_no_file_validation(self, db):
+        """ Test that a validation error is raised rather than any other python exception which would indicate a bug.
+            See https://github.com/ssec-jhu/biospecdb/pull/181
+        """
+        data = SpectralData()
+        with pytest.raises(ValidationError):
+            data.full_clean()
+
     def test_temp_files_deleted(self, mock_data_from_files):
         n_patients = 10
         assert len(SpectralData.objects.all()) == n_patients


biospecdb/apps/uploader/models.py
```diff
@@ -514,7 +514,9 @@ def clean_data_file(self):
 
             The schema is equivalent to `json.dumps(dataclasses.asdict(uploader.io.SpectralData))``.
         """
-        if self.data is None:
+
+        # Note: self.data is a FieldFile and is never None so check is "empty" instead, i.e., self.data.name is None.
+        if not self.data:
```
amitschang, line 519: I'm not sure I understand, the object will return truth-value of False when no file for upload has been set? Is this part of the API that can be relied on? The comment above seems to indicate it could be `self.data.name is None`

jamienoss: >The comment above seems to indicate it could be self.data.name is None

Correct, this is actually how I first changed it, but didn't like that I needed to check against ``name`` as opposed to any other attr. Looking at the django code, using ``__eq__`` over ``is`` is the correct method.

jamienoss: > the object will return truth-value of False when no file for upload has been set?

Correct.

jamienoss: >Is this part of the API that can be relied on?

yes
[src](https://github.com/django/django/blob/main/django/db/models/fields/files.py#L25-L30)
```python
    def __eq__(self, other):
        # Older code may be expecting FileField values to be simple strings.
        # By overriding the == operator, it can remain backwards compatibility.
        if hasattr(other, "name"):
            return self.name == other.name
        return self.name == other
```

amitschang (approved):

PR #183, opened 2023-11-29
Admin tweaks

jamienoss:
The following two changes effectively make inlines "collapsable"/optional.
 * ``min_num = 0``
 * ~``extra = 0``~ PI prefers these to be expanded by default.
 
* Only display the number of given existing observations or an extra equal to the total applicable observables.
* Don't use any inlines for admin pages, do so only for the data admin pages.
* Exclude the following from the data admin pages:
  * Instrument
  * Observable
  * QCAnnotator
  * QCAnnotation


fabbasinejad (approved):

jamienoss:
@fabbasinejad Thanks for the review 👍

PR #184, opened 2023-11-30
Filter only existing bool pie chart data

jamienoss:
``count_bool_observables`` shouldn't fail even if some cols exist but not all.


vkluzner (approved):

jamienoss:
@vkluzner thanks for the quick review 👍

PR #188, opened 2023-12-01
Su only access to admin page

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Restrict-admin-page-to-su-only-d9324ae1bab846b185a3a3340b0b734c?pvs=4)

### DIFF
--- a/biospecdb/apps/uploader/tests/test_uploader_admin.py
+++ b/biospecdb/apps/uploader/tests/test_uploader_admin.py
@@ -77,6 +77,9 @@ class TestAdminPage:
         if url_root == "/data/uploader/" and model in SKIP_MODELS:
             pytest.skip("Model not registered with data admin site.")
 
+        if url_root == "/admin/uploader/" and not user.is_superuser:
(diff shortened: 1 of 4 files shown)


jamienoss:
docker image requires git...

biospecdb/apps/uploader/tests/test_uploader_admin.py
```diff
@@ -77,6 +77,9 @@ def test_admin_pages(self, request, user, url_root, model, action):
         if url_root == "/data/uploader/" and model in SKIP_MODELS:
             pytest.skip("Model not registered with data admin site.")
 
+        if url_root == "/admin/uploader/" and not user.is_superuser:
+            pytest.skip("Admin site restricted to superuser.")
+
```
amitschang, line 82: maybe it is already here, but would it make sense to have a test to ensure the path generates appropriate exception in case user is not superuser?

jamienoss: That's a much better idea 👍

PR #196, opened 2023-12-01
Add data catalog app

jamienoss:
Resolves the following Notion tickets:

 * [Implement pre-canned query model in django](https://www.notion.so/Implement-pre-canned-query-model-in-django-cecb120c88544ad1ab46ddbb2952abac?pvs=4)
 * [Pre-canned dataset catalogue](https://www.notion.so/Pre-canned-dataset-catalogue-9a25f3aaf5e94af68204a99e42b60139?pvs=4)

Implementing this as its own Django app is advantageous as it allows for cleaner separation in permissions, access, DB routing etc, etc.

Screenshot 2023-12-07 at 8 42 04 PM

### TODO:

- [x] Hook up exporter to save dataset
- [x] Add meta data to dataset file
  - [ ] ~Is it a security issue to add SQL query to the resultant downloadable dataset file?~ We don't think so.
- [x] Permissions:
  - [x] Add ``User.is_cataloguser`` & restrict access to this.
  - [x] Hard code readonly perm other than SU.
  - [x] Test permissions
- [x] Untangle/duplicate ``settings.EXPLORER_DATA_EXPORTERS_INCLUDE_DATA_FILES``
- [x] checksums
  - [x] file (inc timestamp so not repeatable - verify download)
  - [x] file contents (repeatable - verify version diffs)
- [x] validations (there's not much to validate)
- [ ] ~auto versioning? (prob not)~ Nope
- [ ] ~inline admin query form?~ Nope, would be better to test the query first.
- [ ] ~Add DOI? (Not sure how this would be possible...)~ Not possible, without making data public.
- [x] file meta:
  * n rows
  * file size
  * spectral data filenames
  * data file checksum
- [ ] ~Encrypt zipfile? Yes, but make it configurable.~ Nope, Python's ``zipfile`` module doesn't support this. See [Notion ticket](https://www.notion.so/Encrypt-zipfiles-69b6048fa1674c5c8678288b88f38a1c?pvs=4).
- [x] zip compression
- [x] fix query django fixture (it's not loading correctly)


biospecdb/apps/catalog/models.py
```diff
@@ -0,0 +1,181 @@
...
+        return Path(str(self).replace('-', '_').replace('.', '_'))
+
+    def clean(self, *args, **kwargs):
+        self.name = self.name or self.query.title
+        self.description = self.description or self.query.description
+        self.sql = self.query.sql
+        super().clean(*args, **kwargs)
+
+        if not self.file:
+            # Execute query for validation.
```
jamienoss, line 102: Yep, we have to unfortunately execute this twice - with a race cond between here and the second run in ``save()``.

jamienoss: Actually, in this instance, we may not have to and could just cache this data. We just can't create the datafiles here, that has to happen in save - anything with side effects must happen in save.

biospecdb/apps/catalog/tests/test_datasets.py
```diff
@@ -0,0 +1,65 @@
...
+        assert dataset.description == query.description
+        assert dataset.sql == query.sql
+        assert dataset.app_version == __version__
+
+    def test_save(self, saved_dataset):
+        assert Dataset.objects.count() == 1
+
+    def test_checksum(self, saved_dataset):
+        checksum = saved_dataset.compute_checksum()
+        assert checksum == saved_dataset.compute_checksum()
```
amitschang, line 40: is this intentionally checking a repeated invocation?

biospecdb/apps/uploader/exporters.py
```diff
@@ -37,15 +39,47 @@ def get_output(self, **kwargs):
...
+                    temp.seek(0)
+                    output = temp
+
+            return (output, (n_rows, data_sha256, spectral_data_filenames)) if return_info else output
+
+        if compression_type is None:
+            compression_type = import_string(settings.ZIP_COMPRESSION)
+
+        if compression_level is None:
+            compression_level = settings.ZIP_COMPRESSION_LEVEL
```
amitschang, line 82: seems a bit confusing that these compression related options are treated differently btwn `include_data_files==True` and otherwise. Any particular reason to construct the zip up above?

biospecdb/apps/uploader/exporters.py
```diff
@@ -67,13 +101,17 @@ def get_file_output(self, **kwargs):
                 # There could be multiple "col_name" (aka "data") columns so flatten first.
                 data_files = df.to_numpy().flatten().tolist()
 
-        if data_files:
-            # Dedupe
-            data_files = set(data_files)
+        if data_files or always_zip:
```
amitschang, line 104: just wondering if makes more sense to have only this block for constructing zip and express the `include_datasets = False` via empty list (iso also having condition above `if not include_datafiles: if always_zip`

amitschang (commented):
some inline comments

biospecdb/apps/catalog/admin.py
```diff
@@ -0,0 +1,105 @@
...
+                "classes": ["collapse"],
+                "fields": ["spectral_data_filenames"],
+            }
+        ),
+    ]
+
+    @admin.display
+    def size(self, obj):
+        if obj.file:
+            return f"{int(obj.file.size / 1e6)} MB"
```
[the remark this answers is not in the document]
jamienoss: or ``int`` -> ``float``. It's the size of the zipped dataset (meta + spec files). For the 181 patients we already have it's ~4MB (compressed). So 🤷

biospecdb/apps/catalog/tests/test_datasets.py
jamienoss: Yep, wanted to make sure it was static.

biospecdb/apps/catalog/admin.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @admin.display
+    def n_spectral_data_files(self, obj):
+        if obj.spectral_data_filenames:
+            return len(obj.spectral_data_filenames)
+
```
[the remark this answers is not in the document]
jamienoss: Huh? No, but I'm not following the question, sorry.

biospecdb/apps/catalog/models.py
```diff
@@ -0,0 +1,180 @@
+import datetime
+import hashlib
+import json
+from pathlib import Path
```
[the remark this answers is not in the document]
jamienoss: I think PEP8 outlines blocking imports by N-paty scope and being alphabetical.

biospecdb/apps/uploader/exporters.py
jamienoss: Good catch. It's confusing because it's wrong and these two conditionals should be hoisted above the zip.

biospecdb/apps/uploader/exporters.py
jamienoss: I'm not sure if I've done exactly this, but have consolidated it.

biospecdb/apps/catalog/admin.py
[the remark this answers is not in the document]
amitschang: what I mean is that `if obj.spectral_data_filenames` resolves to False, presumably it is subsequently the case that `n_` of them is `0`. Now this returns either a number of None for what looks like a count

biospecdb/apps/catalog/tests/test_datasets.py
amitschang: Might be handy to explain the reasoning for that, in case the function later gets touched and potentially dropped

biospecdb/apps/catalog/admin.py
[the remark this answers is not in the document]
jamienoss: Revision made.

biospecdb/apps/catalog/models.py
```diff
@@ -0,0 +1,185 @@
...
+                                 help_text="Number of data rows")
+    data_sha256 = models.CharField(max_length=64,
+                                   editable=False,
+                                   null=False,
+                                   blank=True,
+                                   verbose_name="Data SHA-256",
+                                   help_text="Checksum of data table (not including any spectral data files).",
+                                   validators=[MinLengthValidator(64)])
+    spectral_data_filenames = models.JSONField(null=False,
+                                               default=empty_list,
```
amitschang, line 86: Any particular reason not to `lambda: []` here?

amitschang (approved):
looks good, thanks for the updates. A very minor comment if you like

biospecdb/apps/catalog/models.py
jamienoss: Yep, it needs to be serializable.

PR #203, opened 2023-12-15
List observables by category in inline forms plus other admin tweaks

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Fieldsets-for-observable-catagories-for-inline-patient-form-bf54b01269284cc2b3ac163a699bf236?pvs=4).

This also dynamically uses the correct input widget for ``Observation.observable_value`` based on ``Observation.observable.value_class``.

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -273,6 +273,8 @@ class ObservationInlineForm(forms.ModelForm):
             widget = forms.NumberInput()
         elif value_class is Observable.Types.STR:
             widget = forms.TextInput()
+        else:
+            raise NotImplementedError(f"Dev error: missing widget mapping for type '{value_class}'.")
         return widget
 
     def __init__(self, *args, **kwargs):
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -3,9 +3,10 @@ from inspect import signature
 from django.contrib import admin
 from django.db.models import Q
 from django.db.utils import OperationalError
-import django.forms as forms
+from django import forms
 from nested_admin import NestedStackedInline, NestedTabularInline, NestedModelAdmin
 
+from biospecdb.util import to_bool
 from .models import BioSample, Observable, Instrument, Patient, SpectralData, Observation, UploadedFile, Visit,\
     QCAnnotator, QCAnnotation, Center, get_center, BioSampleType, SpectraMeasurementType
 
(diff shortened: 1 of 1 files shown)


biospecdb/apps/uploader/admin.py
```diff
@@ -176,9 +177,39 @@ def get_queryset(self, request):
 
 
 class ObservationMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        try:
+            # Only auto-populate "global" observables, i.e., those not related to a center (center=null).
+            query = Q(center=None)
```
jamienoss, line 184: This is fortunate as there's no non-hacky way to get access to the user's centers at this point, and this is already hacky enough.

biospecdb/apps/uploader/admin.py
```diff
@@ -231,30 +262,56 @@ def get_queryset(self, request):
...
+        value_class = Observable.Types(value_class)
+        if value_class is Observable.Types.BOOL:
+            widget = forms.CheckboxInput(check_test=to_bool)
+        elif value_class is Observable.Types.FLOAT:
+            widget = forms.NumberInput()
+        elif value_class is Observable.Types.INT:
+            widget = forms.NumberInput()
+        elif value_class is Observable.Types.STR:
+            widget = forms.TextInput()
+        return widget
```
amitschang, line 276: Should we provide either a fallback or (more likely) raise a ValueError in the case we don't match any of the conditions? it would probably be easier to recognize than an undefined symbol type of error.

amitschang (commented):
a comment, also I suspect you will sort out the test coverage check in a subsequent change?

biospecdb/apps/uploader/admin.py
jamienoss: Line 267
```python
value_class = Observable.Types(value_class)
```
Will raise
```python
ValueError: 'blah' is not a valid Observable
```
@amitschang Is that sufficient?

amitschang: Not exactly with respect to my comment. just that if there was a new type outside of (`BOOL`, `FLOAT`, `INT`, `STR`), for whatever reason, this could end up with no error from that line, but an error since widget is undefined. Essentially you are supporting a certain (possiblly not complete) subset of `Types` (specifically `Observable.Types` in this function.

jamienoss: If there were a new type not specified but ``Observables.Types`` this would be incorrect. Additionally, if we're also just talking about additional ``elif`` stmnts that aren't members of ``Observables.Types``, we'd need to add more than just an ``else`` since these would also never make it past ``value_class = Observable.Types(value_class)``.

I'm happy to add an ``else`` as I don't normally like ``elif``s without one, but I also wanna make sure I understand your point.

amitschang: still not sure I get it -- maybe I'm not sure where the input value_class is coming from, but if `Observables.Types` was modified to have `COOL_TYPE` but this function was missed what would happen?

but in any case adding an else is basically the gist of what I was suggesting.

jamienoss: Bingo, now I get you. I thought you meant if one wasn't added to ``Types`` but just as an ``elif`` here.

Code change on its way...

jamienoss:
@amitschang Hey, I'm not quite sure what's going on with the codecov diff because when I run the tests locally, only 8 lines are missing and not 13 as [reported here](https://github.com/ssec-jhu/biospecdb/pull/203/checks?check_run_id=20318719967).
Screenshot 2024-01-09 at 4 31 19 PM

I'm gonna investigate this in another PR, so have re-requested your review, please.

amitschang (approved):

jamienoss:
@amitschang Thanks for the review.

@glemson Are you still looking at this or am I ok to merge?

glemson:
@jamienoss let me finish my review this morning

PR #210, opened 2024-01-18
Gunicorn, nginx, and (semi) prd deployment

jamienoss:
## Deployment expectations:
This is not necessarily _the_ final deployment, we're waiting for definition of which cloud service provider the PI would like to use. Even then, we'll probably only whip up a working example/foundation for them. With that in mind, the docker-compose script is similar in that context, however, we can now at least provide _a_ mechanism for standing up a prd service if need be.

## Changes:
* Use ``gunicorn`` in prd deployment.
* Use ``nginx`` in front of gunicorn for all traffic.
* Serve static files (e.g., .css) and media files (e.g., downloading data files) from ``nginx``.
* Configure ``docker-compose`` for prd deployment.

## TODO?:
* Build & deploy using ``docker-compose`` as part of github CI?
* what to do about github package for docker image?

## Punt:
* Need domain name for:
  * ``biospecdb.settings.prd.ALLOWED_HOSTS``
  * ``nginx/nginx.conf:server_name``
* See #211 for SSL/https: we need certificates. Changes needed for:
  * ``config/gunicorn/prd.py``
  * ``nginx/nginx.conf``
  * ``biospecdb/settings/prd.py``

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,15 +41,21 @@ For additional cmds see the [Conda cheat-sheet](https://docs.conda.io/projects/c
 
 ### Run:
 
-  #### using Docker:
+  #### using Docker (for production):
   * Download & install Docker - see [Docker install docs](https://docs.docker.com/get-docker/).
   * ``cd`` into repo dir.
-  * Build and run with ``DJANGO_SUPERUSER_PASSWORD=admin docker compose up``
+  * Setup env vars:
+    * ``export DJANGO_SUPERUSER_PASSWORD=admin``
+    * ``source scripts/prd/gen_secret_key.sh``
+  * Build and run with ``docker compose up``, add ``-d`` to run in the background.
   * Alternatively, images can be pulled from ``ghcr.io/ssec-jhu/`` e.g., ``docker pull ghcr.io/ssec-jhu/base-template:pr-1``.
+  * The site can then be accessed using any browser from ``http://localhost``
 
-  #### with Python ecosystem:
+  #### with Python ecosystem (for development):
   * Follow the above [Build with Python ecosystem instructions](#with-python-ecosystem).
-  * Run ``uvicorn biospecdb.asgi:application --host 0.0.0.0 --port 8000``. _NOTE: This is just an example and is obviously application dependent._
+  * For a completely fresh start and rebuild of the database: ``./scripts/dev/rebuild_db.sh``.
+  * Run ``DJANGO_SETTINGS_MODULE=biospecdb.settings.dev python manage.py runserver 0.0.0.0:8000``
+  * The site can then be accessed using any browser from ``http://localhost:8000``
 
 
 ### Custom Deployment Settings:
@@ -124,7 +130,7 @@ population of the ``QCAnnotation`` table is configurable and is described below.
 
 #### Settings:
 
-The following QC annotator settings are available in ``biospecdb.settings``:
+The following QC annotator settings are available in ``biospecdb.settings.base``:
 
  * ``AUTO_ANNOTATE``: If ``True`` and if default annotators exist in the DB, they will be automatically run upon
                       adding/updating ``SpectralData`` entries. _(Default: True)_
--- a/biospecdb/settings/base.py
+++ b/biospecdb/settings/base.py
@@ -15,25 +15,15 @@ import os
 import sys
 
 
-PROJECT_ROOT = os.path.dirname(__file__)
+PROJECT_ROOT = Path(os.path.dirname(__file__)).parent
 sys.path.insert(0, os.path.join(PROJECT_ROOT, 'apps'))
 
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
-BASE_DIR = Path(__file__).resolve().parent.parent
-
+BASE_DIR = Path(__file__).resolve().parent.parent.parent
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/4.1/howto/deployment/checklist/
-
-# SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = 'django-insecure-9ppt7g3y2ds5+nig07x(#8^th-olh5u=kr_tiqs$2*h(u!y&^m'
-
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
-
-ALLOWED_HOSTS = ['*']
-
 
 # Application definition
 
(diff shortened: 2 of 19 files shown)


biospecdb/settings/base.py
```diff
@@ -137,7 +127,9 @@
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/4.1/howto/static-files/
 
-STATIC_URL = 'static/'
+STATIC_ROOT = BASE_DIR / "static/"
+STATIC_URL = '/static/'
+STATICFILES_DIRS = [os.path.join(BASE_DIR, 'biospecdb/apps/uploader/templates/static')]
```
amitschang, line 132: any reason not `[BASE_DIR / 'biospecdb/apps/uploader/templates/static']`? go all in with pathlib!

config/gunicorn/prd.py
```diff
@@ -0,0 +1,51 @@
...
+import multiprocessing
+
+
+# Django WSGI application path in pattern MODULE_NAME:VARIABLE_NAME
+wsgi_app = "biospecdb.wsgi:application"
+
+# The socket to bind
+bind = "0.0.0.0:8000"
+
+# The granularity of Error log outputs
```
amitschang, line 11: similar warning applies for debug logs in prod from gunicorn?

config/gunicorn/prd.py
```diff
@@ -0,0 +1,51 @@
...
+loglevel = "info"
+
+# The number of worker processes for handling requests.
+# Note: this is multiprocessing and not multithreading, because the GIL.
+# Note: Keep in mind that sqlite blocks on all connections and has no concurrency!
```
amitschang, line 16: isn't that only for writes?

docker-compose.yml
```diff
@@ -35,24 +38,60 @@ services:
...
+    deploy:
+      mode: global
+      restart_policy:
+        condition: on-failure
+        delay: 5s
+        max_attempts: 3
+        window: 10s
+    container_name: bsr-nginx
+    ports:
+      - "80:80"
```
amitschang, line 77: I see here, but I would guess this is not typical for a production installation. My guess is usually there would be an SSL-terminating proxy at a layer above this, and in such a case then that could direct traffic to the gunicorn process, and perhaps not much need for an nginx config within this project? Or is this to serve a particular setup?

docker-compose.yml
```diff
@@ -14,11 +14,14 @@ services:
       - bulk-upload-files:/app/raw_data
       - spectaldata-files:/app/datasets
     environment:
-      DJANGO_SETTINGS_MODULE: biospecdb.settings
+      DJANGO_SETTINGS_MODULE: biospecdb.settings.prd
```
amitschang, line 17: should there be another one of these for development (where perhaps docker compose is actually most useful)? I might also lack some context for exactly how the deployment is planned (but then also some of the prod deployment details for the project may not necessarily belong in this repo)

amitschang (commented):
I have some comments, but also I think I might be missing something about the details of a planned production deployment. And if there is such a planned deployment, I wonder if we should encode the details of that separately from this project repo?

docker-compose.yml
jamienoss: nginx _is_ the proxy layer directing traffic to gunicorn.

It's also the defacto method for serving static files and media files instead of attempting to use gunicorn to do this.

docker-compose.yml
jamienoss: I wondered that, but decided that there's no reason that we can't use the prd settings, other than maybe wanting to change the log level but I don't think warrants the duplication.

>but then also some of the prod deployment details for the project may not necessarily belong in this repo

Using another repo is the only way I'm familiar with, especially since this repo is public, however, I'm starting to think that it may not be necessary. At least not in this PR until we need cloud deployment config.

config/gunicorn/prd.py
jamienoss: Doh, yep.

config/gunicorn/prd.py
jamienoss: Sorry, not following?

README.md
```diff
@@ -41,15 +41,21 @@ For additional cmds see the [Conda cheat-sheet](https://docs.conda.io/projects/c
...
+  #### using Docker (for production):
   * Download & install Docker - see [Docker install docs](https://docs.docker.com/get-docker/).
   * ``cd`` into repo dir.
-  * Build and run with ``DJANGO_SUPERUSER_PASSWORD=admin docker compose up``
+  * Setup env vars:
+    * ``export DJANGO_SUPERUSER_PASSWORD=admin``
+    * ``source scripts/prd/gen_secret_key.sh``
+  * Build and run with ``docker compose up``, add ``-d`` to run in the background.
   * Alternatively, images can be pulled from ``ghcr.io/ssec-jhu/`` e.g., ``docker pull ghcr.io/ssec-jhu/base-template:pr-1``.
+  * The site can then be accessed using any browser from ``http://localhost``
```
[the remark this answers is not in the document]
jamienoss: nginx listens on 80 as a proxy and then traffics requests to and from gunicorn over the internal container network with port 8000 exposed only internally and gunicorn listening on it.

biospecdb/settings/base.py
```diff
@@ -15,26 +15,16 @@
 import sys
 
 
-PROJECT_ROOT = os.path.dirname(__file__)
+PROJECT_ROOT = Path(os.path.dirname(__file__)).parent
```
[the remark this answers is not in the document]
jamienoss: This stuff was all just copied, but yep, I'll look into unifying these.

docker-compose.yml
amitschang: I indeed missed that it is serving the static content. Even in this case I think many scenarios will have something above here as well, and port 80 would not be necessary - and the only reason I'm hung up on it is that it is privileged port and I think the docker-compose stuff here is particularly useful for development and likely more so than many production environments.

config/gunicorn/prd.py
amitschang: Sorry, just meant that for the application config under logging you have `SECURITY WARNING: don't run with debug turned on in production!`, might this also apply to the gunicorn logs?

jamienoss: I think you're referring to the following in ``biospecdb.settings.prd``
```python
# SECURITY WARNING: don't run with debug turned on in production!
DEBUG = False
```
This is a [Django setting](https://docs.djangoproject.com/en/5.0/ref/settings/#debug) and not the log level.

amitschang (approved):

PR #216, opened 2024-01-25
Improvements to Visit.previous_visit

jamienoss:
Replacement for #212

Previous visit hasn't been removed from the model or the forms, it has been slightly hidden from the forms as a collapsable formset.

Collapsed (default):
Screenshot 2024-01-25 at 2 51 09 PM

expanded:
Screenshot 2024-01-25 at 2 51 18 PM

``settings.AUTO_FIND_PREVIOUS_VISIT`` has been turned off as the default functionality. Sequential age validation has been added when this flag is ``True`` to prevent incorrect results when visits are retrospectively added, i.e., out of order in real-time - can't add a visit with age less than the last visit. Though, [Make patient_age an observable](https://www.notion.so/Make-patient_age-an-observable-67b934b46bad4bbaaa89fd848bd18bfd?pvs=4) may perhaps destroy this functionality entirely.

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -473,18 +473,25 @@ class BioSampleInline(BioSampleMixin, RestrictedByCenterMixin, NestedStackedInli
         return 0 if obj and obj.pk and obj.bio_sample.count() else self.extra
 
 
-class VisitAdminForm(forms.ModelForm):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        if self.instance.patient_id:
-            self.fields["previous_visit"].queryset = Visit.objects.filter(patient=self.instance.patient_id)
-
-
 class VisitAdminMixin:
-    form = VisitAdminForm
     readonly_fields = ["created_at", "updated_at"]  # TODO: Might need specific user group.
     ordering = ("-updated_at",)
-    fields = ("patient", "patient_age", "previous_visit")
+
+    fieldsets = [
+        (
+            None,
+            {
+                "fields": ["patient", "patient_age"]
+            }
+        ),
+        (
+            "Advanced",
+            {
+                "classes": ["collapse"],
+                "fields": ["previous_visit"],
+            }
+        ),
+    ]
 
     @admin.display
     def patient_id(self, obj):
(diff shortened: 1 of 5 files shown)


biospecdb/apps/uploader/models.py
```diff
@@ -281,7 +282,7 @@ def center(self):
         return self.patient.center
 
     def __str__(self):
-        return f"patient:{self.patient.short_id()}_visit:{self.visit_number}"
+        return f"patient:{self.patient}_visit:{self.visit_number}"
```
jamienoss, line 285: Slight smuggle, but hey.

biospecdb/apps/uploader/models.py
```diff
@@ -228,7 +228,8 @@ def clean(self):
             raise ValidationError(_("Previous visits do not belong to this patient!"), code="invalid")
 
         # Validate visits are entered ordered by age.
-        if self.previous_visit is not None and (self.patient_age < self.previous_visit.patient_age):
+        if settings.AUTO_FIND_PREVIOUS_VISIT and (self.previous_visit is not None) and \
```
jamienoss, line 231: We should be allowed to add visits out of order - at least when ``settings.AUTO_FIND_PREVIOUS_VISIT=False``.

biospecdb/apps/uploader/admin.py
```diff
@@ -473,26 +473,33 @@ def get_extra(self, request, obj=None, **kwargs):
...
+    ]
 
     @admin.display
     def patient_id(self, obj):
         return obj.patient.patient_id
 
     @admin.display
     def visit_count(self, obj):
-        return obj.visit_number
+        return Visit.objects.filter(patient=obj.patient).count()
```
jamienoss, line 502: This was probably a more accurate method to use in the first place.

jamienoss:
@vkluzner & @glemson sorry for running us in circles on this one.

vkluzner (dismissed):

vkluzner (approved):

biospecdb/apps/uploader/models.py
jamienoss: Nope, this isn't quite correct either as it doesn't validate when ``previous_visit`` was manually entered by the user. In that scenario it should still be validated.

jamienoss: In fact, it should always be validated - we should only be able to add visits out of order when ``settings.AUTO_FIND_PREVIOUS_VISIT=False`` and we don't manually enter a previous visit. When a previous visit is entered, manually or automatically, it should always be validated.

jamienoss: Fixed in #223

PR #217, opened 2024-01-25
Add Visit.days_observed and re-handle Observation.days_observed & Observation.severity

jamienoss:
Replacement for #213


biospecdb/apps/uploader/admin.py
```diff
@@ -238,7 +238,9 @@ def __init__(self, *args, **kwargs):
             "More details",
             {
                 "classes": ["collapse"],
-                "fields": [("created_at", "updated_at")]
+                "fields": ["days_observed",
+                           "severity",
```
vkluzner, line 242: Should `severity` be related to some specific disease? Or `ObservationMixin` is the one it's related to?

biospecdb/apps/uploader/migrations/0001_initial.py
```diff
@@ -182,6 +182,7 @@ class Migration(migrations.Migration):
                 ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('created_at', models.DateTimeField(auto_now_add=True)),
                 ('updated_at', models.DateTimeField(auto_now=True)),
+                ('days_observed', models.IntegerField(blank=True, default=None, help_text='Applies to all observations unless otherwise stated', null=True, validators=[django.core.validators.MinValueValidator(0)], verbose_name='Days of symptoms onset')),
```
vkluzner, line 185: I think "help_text='Applies to all observations unless otherwise stated'" is confusive in the same way as severity. Is there any way to connect it to specific `observable`? I think we discussed the similar issue in the framework of front-end form building (which is cancelled now).

biospecdb/apps/uploader/models.py
```diff
@@ -198,14 +198,22 @@ class Meta:
     previous_visit = models.ForeignKey("self", default=None, blank=True, null=True, on_delete=models.SET_NULL,
                                        related_name="next_visit")
 
+    days_observed = models.IntegerField(default=None,
```
vkluzner, line 201: Same - shouldn't it better belong to `Observable` and not `Visit`?

vkluzner (commented):
Comments are in the body of the code.

biospecdb/apps/uploader/models.py
jamienoss: Do you mean ``Observation``?

jamienoss:
@vkluzner are we talking about the fact that ``severity`` and ``days_observed`` don't necessarily apply to all observable categories, e.g., test, bloodwork, patient_info etc?

jamienoss:
BTW, the PI said that we can nuke ``Observation.severity``, however, they would like to keep ``Observation.days_observed`` for future use.

So maybe do just that, but perhaps tweaking the verbose name and help text for ``days_observed`` will mitigate @vkluzner 's concerns?

jamienoss:
@vkluzner I've added two more commits to:
  * nuke ``Observation.severity``
  * remove "symptoms" from ``Observation.days_observed`` verbiage.

vkluzner:
> BTW, the PI said that we can nuke `Observation.severity`, however, they would like to keep `Observation.days_observed` for future use.
> 
> So maybe do just that, but perhaps tweaking the verbose name and help text for `days_observed` will mitigate @vkluzner 's concerns?

It's OK per `Observation`. My concern was about the `Visit`, which (as far as I remember) can include multiple `Observations`.

vkluzner:
> @vkluzner I've added two more commits to:
> 
> * nuke `Observation.severity`
> * remove "symptoms" from `Observation.days_observed` verbiage.
> 
> What do you think? I really don't want to have to special case `Observables` and `Observations` such that their fields differ as a function of `Observable.Category` - I feel like it would be defeating the point.

Sounds good. I was mostly concerned about `Visit` (see my comment below).

vkluzner (approved):

jamienoss:
>My concern was about the Visit, which (as far as I remember) can include multiple Observations.

So this was the main point - the PI doesn't think that the clinicians will add "days observed" for each and every observation so it needs to go somewhere more global - the only suitable model being ``Visit``. Additionally, the bulk data upload contains only a single value for ``dyas_observed`` which, before this PR, we auto-populated each observation with, however, it would be better to associate this only with the visit.

That being said, the PI also wanted to keep ``Observation.days_observed`` as they will want to create their own trial in the future and they requested that it stay. @glemson then came up with the idea for us to add it to ``Visit`` instead of auto-populating all observations so that we can semantically keep both options rather than nuking ``Observation.days_observed``.

vkluzner:
> > My concern was about the Visit, which (as far as I remember) can include multiple Observations.
> 
> So this was the main point - the PI doesn't think that the clinicians will add "days observed" for each and every observation so it needs to go somewhere more global - the only suitable model being `Visit`. Additionally, the bulk data upload contains only a single value for `dyas_observed` which, before this PR, we auto-populated each observation with, however, it would be better to associate this only with the visit.
> 
> That being said, the PI also wanted to keep `Observation.days_observed` as they will want to create their own trial in the future and they requested that it stay. @glemson then came up with the idea for us to add it to `Visit` instead of auto-populating all observations so that we can semantically keep both options rather than nuking `Observation.days_observed`.

I believe it's OK once PI is happy with it.

PR #218, opened 2024-01-26
Delete media files upon deletion from db

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Delete-uploaded-files-when-model-instances-are-deleted-79fc145181094581aba66e92ec171a8f?pvs=4)

### DIFF
--- a/biospecdb/apps/uploader/tests/test_models.py
+++ b/biospecdb/apps/uploader/tests/test_models.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 from uuid import uuid4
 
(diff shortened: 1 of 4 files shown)


biospecdb/apps/uploader/tests/test_models.py
```diff
@@ -335,6 +336,14 @@ def test_clean(self, centers, instruments, ext, bio_sample_types, spectra_measur
         cleaned_data = spectral_data.get_spectral_data()
         assert cleaned_data == data
 
+    def test_deletion(self, mock_data_from_files):
+        spectral_data = SpectralData.objects.all()
```
vkluzner, line 340: Is it worth adding meta_data definition also for the test needs? (See lines 506-513)

vkluzner (commented):
Approved, but see the comment about meta_data

vkluzner (approved):

biospecdb/apps/uploader/tests/test_models.py
jamienoss: We do, under ``TestUploadedFile``.

PR #219, opened 2024-01-26
Optionally delete all entries created from bulk uploads upon uploaded file deletion

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Optionally-delete-all-entries-created-from-bulk-uploads-upon-uploaded-file-deletion-480f936d0eb24a56beac38423f767864?pvs=4)

~When deleting a bulk uploaded data file, also delete all DB entries created from the bulk data upload. Note: This won't delete the patient entry itself, just the visit and all other data related to the visit. This is because there's no way to know whether the patient previously existed even if it had no associated visit.~

~This functionality is conditional on ``settings.DELETE_VISITS_WHEN_DELETING_BULK_UPLOAD``~

I'm gonna table this for now - I think the better way to go (if at all) is to either conditionally set ``Visit.src_file(on_delete=models.CASCADE if settings.DELETE_VISITS_WHEN_DELETING_BULK_UPLOAD else models.SET_NULL)`` or, better still, add a custom admin action to explicitly cascade delete with its own perms etc.

### UPDATE:
I couldn't find a suitable path to hook into or reuse Django's deletion methods to force a cascade.
Additionally, conditionally setting ``on_delete`` isn't robust as it requires more than just changing that setting but also requires migrations to be run.
I'm gonna just ask the PI if they want these to cascade and then just set it in the model - job done.


jamienoss:
Looking into failing tests as these pass locally...

vkluzner (dismissed):

jamienoss:
Merge conflicts resolved and fixed - see additional commit.

@glemson are you wanting to review this before merging?

amitschang:
@jamienoss I will look, sorry for delay

jamienoss:
When deleting the uploaded bulk data file all objs to be deleted should be list in the conformation warning page, which with the current code, they aren't.
Screenshot 2024-01-31 at 2 45 56 PM

jamienoss:
FYI, I'm rethinking this entirely, as this is effectively just a cascading delete that I'm manually trying to re-create when we don't need to.

PR #221, opened 2024-01-26
Use GitHub actions cache for pip deps

jamienoss:


amitschang (approved):
never used this feature, but looks good from here

PR #222, opened 2024-01-27
Update missing catalog migration

jamienoss:


jamienoss:
Will admin merge as soon as tests pass?

PR #225, opened 2024-01-28
Semantic changes to Observable.center and form auto inlining

jamienoss:
Resolves [Notion ticket](https://www.notion.so/changes-to-observable-center-semantics-76915aa602fd43bfb769aaccf50a59b3?pvs=4)

Changes:
* add help text for center functionality
*  ~“———” → “all”~ Nope, for many2many fields there is no "null" choice. Instead, all choices remain unselected, since this it uses a multi-select widget.
*  Make ``Observable.center`` a many-to-many field rather than a many-to-one.

TODO:
- [x] Merge #228 
- [x] check bulk upload center validation
- [x] check queryset correctness of all ``Observable.center``

Punt:
 * center local defaults (not just global). Implemented by adding ``Observable.default`` which is a  many-to-many field to ``User.center`` indicating for which centers observations should be auto-inlined with populated observables. As is/was  ``Observable.center=None`` indicated whether an observable be auto-inlined. The technical difficulty with implementing this is that the existing auto-inline-population is hacky at best and is done at admin and form instantiation which obviously doesn't have access to user request info, i.e., ``request.user.center``. This info is needed to facilitate the observable filtering by user center. Instead, this entire implementation needs to be properly implemented within the Django framework that does grant dev access to the user request, e.g., ``admin.ModelAdmin.get_inlines(self, request, obj)`` or ``admin.ModelAdmin.get_formsets_with_inlines(self, request, obj=None)``. For this reason, this functionality has been deferred to #249 and ``Observable.default`` will be reverted in this PR such that at least the other changes can be merged whilst (and if at all) I can figure out the alternative.


biospecdb/apps/uploader/models.py
```diff
@@ -427,10 +440,10 @@ def clean(self):
         """ Model validation. """
         super().clean()
 
-        if self.observable.center and (self.observable.center != self.visit.patient.center):
-            raise ValidationError(_("Patient observation observable category must belong to patient center: "
-                                    "'%(c1)s' != '%(c2)s'"),
-                                  params=dict(c1=self.observable.center, c2=self.visit.patient.center))
+        # Note: global observables have no observable.center.
+        if self.observable.center.count() and (self.visit.patient.center not in self.observable.center.all()):
```
vkluzner, line 444: `self.observable.center.count() ` - Do you really need this check? If it's 0, shouldn't the second part of condition be 0 by definition?

jamienoss: > Do you really need this check?

Yes, very much so because we're checking ``if A is NOT in B`` because everything and anything is simultaneously NOT in an empty set.
```python
A = []
if None not in A:
  raise Exception
```
The above will raise and the following will not.
```python
A = []
if A and (None not in A):
  raise Exception
```

PR #226, opened 2024-01-29
Fix previous_visit formfield_for_foreignkey regression from #216

jamienoss:
Resolves regression from #216

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -526,7 +526,11 @@ class VisitAdminMixin:
                 return field
 
             # Parse the url_name and retrieve the model for the above object_id.
-            app, model, action = request.resolver_match.url_name.split('_')
+            try:
+                app, model, action = request.resolver_match.url_name.split('_')
+            except ValueError:  # too many values to unpack.
+                return field
+
             try:
                 model = apps.get_model(app, model)
             except LookupError:
@@ -538,11 +542,13 @@ class VisitAdminMixin:
             try:
                 # Limit the QS for previous_visit.
                 if model is Patient:
-                    # Limit to all visits belonging to this patient (inc self).
+                    # Limit to all visits belonging to this patient (inc self - unfortunatley. It would be better to
+                    # exclude (See below), however, there is no visit-self when looking at the patient level).
                     patient = Patient.objects.get(pk=object_id)
                     field.queryset = Visit.objects.filter(patient=patient)
                 elif model is Visit:
-                    # Limit to all visits belonging to this patient (exc self).
+                    # Limit to all visits belonging to this patient (exc self - since a self-referential previous_visit
+                    # doesn't make much sense).
                     visit = Visit.objects.get(id=object_id)
                     patient = visit.patient
                     field.queryset = Visit.objects.filter(patient=patient).exclude(pk=object_id)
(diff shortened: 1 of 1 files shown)


biospecdb/apps/uploader/admin.py
```diff
@@ -517,21 +519,38 @@ def formfield_for_foreignkey(self, db_field, request, **kwargs):
         """ Limit previous_visit to user's center (super's functionality) and admin category. """
         field = super().formfield_for_foreignkey(db_field, request, **kwargs)
         if db_field.name == "previous_visit":
-            visit_pk = request.resolver_match.kwargs.get("object_id", None)
```
jamienoss, line 520: This isn't possible as we're not sure, at this point, which model this is an ID for.

biospecdb/apps/uploader/admin.py
```diff
@@ -517,21 +519,38 @@ def formfield_for_foreignkey(self, db_field, request, **kwargs):
...
-            patient = visit.patient if visit else None
 
-            # Note: Our forms aren't dynamic and thus this still won't solve the situation where a new visit is
+            # Get the object ID - not yet sure which model the ID belongs to.
+            object_id = request.resolver_match.kwargs.get("object_id", None)
+            if not object_id:
+                return field
+
+            # Parse the url_name and retrieve the model for the above object_id.
+            app, model, action = request.resolver_match.url_name.split('_')
```
amitschang, line 529: I don't know enough about this to tell, but is it guaranteed that the `url_name` (which from the docs seems like "The name of the URL pattern that matches the URL.") is a triple of that representation? i.e. should this also be wrapped and return `field` if it doesn't match that representation?

biospecdb/apps/uploader/admin.py
```diff
@@ -517,21 +519,38 @@ def formfield_for_foreignkey(self, db_field, request, **kwargs):
...
-                if request.user.is_superuser:
-                    field.queryset = Visit.objects.all()
+            try:
+                # Limit the QS for previous_visit.
+                if model is Patient:
+                    # Limit to all visits belonging to this patient (inc self).
+                    patient = Patient.objects.get(pk=object_id)
+                    field.queryset = Visit.objects.filter(patient=patient)
+                elif model is Visit:
+                    # Limit to all visits belonging to this patient (exc self).
```
amitschang, line 545: can see that, perhaps best to use comment to clarify why that is necessary

biospecdb/apps/uploader/admin.py
jamienoss: It should be, as long as neither uses '_' in the name itself, which it could in the future, so good call.

biospecdb/apps/uploader/admin.py
```diff
@@ -517,21 +519,44 @@ def formfield_for_foreignkey(self, db_field, request, **kwargs):
...
-            else:
-                if request.user.is_superuser:
-                    field.queryset = Visit.objects.all()
+            try:
+                # Limit the QS for previous_visit.
+                if model is Patient:
+                    # Limit to all visits belonging to this patient (inc self - unfortunatley. It would be better to
+                    # exclude (See below), however, there is no visit-self when looking at the patient level).
+                    patient = Patient.objects.get(pk=object_id)
+                    field.queryset = Visit.objects.filter(patient=patient)
```
jamienoss, line 548: FYI, this (and one below) will be addressed post-merge by #226.

amitschang (approved):
Looks good, thanks!

PR #228, opened 2024-02-01
Fix formfield_for_foreignkey inheritance call order clobbering

jamienoss:
Changes:
 * Model admin ``formfield_for_foreignkey`` methods were clobbering their supers instead of filtering on them. This PR fixes that.
 * Tighten object visibility to the user's center across all fields.

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -19,12 +19,12 @@ class RestrictedByCenterMixin:
         user_center = request.user.center if request.user else None
 
         if (not user_center) or (obj is None):
-            # Those without centers "own" all.
-            return True  # security risk!?
+            return False  # Strict security.
 
         obj_center = get_center(obj)
         if not obj_center:
-            # Objects without centers are "owned" by all.
+            # This func is called for new obj forms where obj.center obviously hasn't been set yet. In this scenario
+            # limited visibilty is defered to self.formfield_for_foreignkey.
             return True
 
         return obj_center == user_center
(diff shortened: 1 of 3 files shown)


biospecdb/apps/uploader/admin.py
```diff
@@ -68,15 +68,48 @@ def formfield_for_foreignkey(self, db_field, request, **kwargs):
...
+            return field
+
+        # User.center can't be Null so this isn't actually possible, but just in case return an empty queryset.
+        if not request.user.center:
+            field.queryset = field.queryset.none()
+            return field
+
+        center = Center.objects.get(pk=request.user.center.pk)
+
+        if db_field.name == "center":
```
jamienoss, line 83: These could be added to each of the respective model admins, however, for complete and consistent coverage, it's better to group them all here.

biospecdb/apps/uploader/admin.py
```diff
@@ -304,7 +337,7 @@ def formfield_for_foreignkey(self, db_field, request, **kwargs):
         """ Limit observable to user's center (super's functionality) and admin category. """
         field = super().formfield_for_foreignkey(db_field, request, **kwargs)
         if db_field.name == "observable":
-            field.queryset = Observable.objects.filter(category=self.verbose_name.upper())
+            field.queryset = field.queryset.filter(category=self.verbose_name.upper())
```
jamienoss, line 340: Example of how the querysets were previously getting clobbered.

jamienoss:
Anyone gonna review this please, pretty please, with a cherry on top?

vkluzner (approved):
Approved it with a comment in the body.

biospecdb/apps/uploader/admin.py
```diff
@@ -68,15 +75,52 @@ def formfield_for_foreignkey(self, db_field, request, **kwargs):
...
+            return field
+
+        # User.center can't be Null so this isn't actually possible, but just in case return an empty queryset.
+        try:
+            if not request.user.center:
+                field.queryset = field.queryset.none()
+                return field
+        except User.center.RelatedObjectDoesNotExist:
+            field.queryset = field.queryset.none()
+            return field
```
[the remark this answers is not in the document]
jamienoss: ``RelatedObjectDoesNotExist`` is an ``Exception`` and each model has its own and these have to be caught (or their model base). Either way, it can't be bundled in with the ``if`` since it's an exception - it can only be handled by ``try:except``.

jamienoss:
@vkluzner Thanks for the review.

PR #234, opened 2024-02-02
Turn codecov back on for dependabot PRs

jamienoss:


jamienoss:
Will admin merge once tests pass.

PR #236, opened 2024-02-07
Only auto-populate Observation.observable for inline forms

jamienoss:
As it stood, the auto-population of ``Observation.observable`` when adding a new observation (not inlined) gave a random observable, but more importantly loaded the widget for that observable's value class and isn't editable after the fact - basically, making it unusable.

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -212,18 +212,6 @@ class ObservableAdmin(RestrictedByCenterMixin, NestedModelAdmin):
 
 
 class ObservationMixin:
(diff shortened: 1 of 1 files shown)


biospecdb/apps/uploader/admin.py
```diff
@@ -291,6 +279,18 @@ def __init__(self, *args, **kwargs):
 
 
 class ObservationInline(ObservationMixin, RestrictedByCenterMixin, NestedTabularInline):
+    def __init__(self, *args, **kwargs):
```
jamienoss, line 282: No code was changed - just its location.

PR #239, opened 2024-02-07
Gender observable

jamienoss:
Resolves:
 * [Make-gender-an-observable](https://www.notion.so/Make-gender-an-observable-6d3fa720ea07464e8ed9b6cf2ea9deab?pvs=4)
 * [Model-disease-observable-value-types](https://www.notion.so/Model-disease-observable-value-types-b35c8728b2634d83affa623284cfccf1?pvs=4)

## Changes:
* Add ``Observable.validator``
* Add ``Observable.value_choices``
* Remove ``Pateint.gender``
* Add ``Observable(name="gender")``

## Limitations:
Within our current framework, there are no dynamic forms, so if you add a new observation of an observable with choices, unfortunately, these choices are not displayed in the input field. However, this is only true when adding an individual observation. Inline observations have their observable pre-populated and therefor the correct widget can be loaded which will include given choices (if any). See [Add-javascript-and-onchange-attr-to-observation-widget](https://www.notion.so/Add-javascript-and-onchange-attr-to-observation-widget-06dbe50b3aed4867ab98452076b41e73?pvs=4).

(with choices widget)
Screenshot 2024-02-07 at 3 16 44 PM

(without choices widget)
Screenshot 2024-02-07 at 3 17 48 PM

This will also mean that existing observations can not simultaneously edit both an observation's observable and value, because the value widget will belong to the previous observable. There is scope for this to work as is when editing to a similar observable.


jamienoss:
@glemson see "limitations" in the description.

glemson:
So is it possible to define a new Observable and indicate its values should come from a defined list?

biospecdb/apps/uploader/admin.py
```diff
@@ -538,7 +542,10 @@ def visit_count(self, obj):
 
     @admin.display
     def gender(self, obj):
-        return obj.patient.gender
+        try:
+            return obj.observation.get(observable__name="gender").observable_value
```
glemson, line 546: would not be stable against name change. Can't do much about that I suppose. 
What was reason not to have gender as hardcoded field, whereas we have it as a hardcoded method here?
Also, since it is used multiple times might it not be better to use some global constant (?) to store this name?

biospecdb/apps/uploader/fixtures/observables.json
```diff
@@ -297,5 +297,19 @@
...
+  {
+    "model": "uploader.Observable",
+    "pk": 24,
+    "fields": {
+      "category": "PATIENT_INFO",
+      "name": "gender",
+      "description": "Patient's gender at time of visit.",
+      "alias": "Gender",
+      "value_class": "STR",
+      "value_choices": "Prefer not to state, non-binary, Agender/I don’t identify with any gender, transgender woman, transgender man, woman, man",
```
glemson, line 310: Is order something we should consider?

jamienoss:
@glemson 
>So is it possible to define a new Observable and indicate its values should come from a defined list?

Yes.

>I see it could be tricky allowing someone to edit an observable, say remove values from the list, if someone else already assigned that value in an observation.

Observables can only be added/edited via the admin page, not the data page, and for those with explicit perms - which should only be the admin/superuser.

The issue here isn't with observables but with an observation of an observable that defines choices. See screenshots above, and I'll demonstrate in person tomorrow.

The issue can be remedied with some javascript, but if we do that (at all) it'll be beyond the scope of this PR.

biospecdb/apps/uploader/models.py
```diff
@@ -310,6 +302,16 @@ def __str__(self):
         return f"patient:{self.patient}_visit:{self.visit_number}"
 
 
+def validate_import(value):
```
glemson, line 305: could you add some comment what this function is for?

biospecdb/apps/uploader/models.py
```diff
@@ -379,7 +409,10 @@ class Meta:
                                         help_text="Supersedes Visit.days_observed")
 
     # Str format for actual type/class spec'd by Observable.value_class.
-    observable_value = models.CharField(blank=True, null=True, default='', max_length=128)
+    observable_value = models.CharField(blank=True,
+                                        null=True,
+                                        default='',
+                                        max_length=128)
```
glemson, line 415: sure this is long enough? For all conceivable observables?

glemson (commented):
few questions. I think I'd like "gender" to be a constant or whatever is suitably pythonic..

biospecdb/apps/uploader/admin.py
jamienoss: >What was reason not to have gender as hardcoded field

Because you, I, and the PI decided a few tag-ups ago that all the patient info should go under the patient info observable as an observation such that all such similar info maintained a single interface. That interface being the ``Observable(category="PATIENT_INFO")``. If we made gender a field (which it obviously was), we'd have to make the others fields also, but instead we decided to go the other route.

jamienoss: Since ``Observable(category="PATIENT_INFO")`` exists, we decided that a user should query similar info on this observable and not have to special case a filter on gender.

biospecdb/apps/uploader/fixtures/observables.json
jamienoss: I did, and explicitly put it in this order.

Also, I meant to mention that I obtain these from [reimaginegender.org/insights/gender-and-forms](https://reimaginegender.org/insights/gender-and-forms?gad_source=1&gclid=CjwKCAiA8YyuBhBSEiwA5R3-E3Iu1we-YcSGNcCuxVxRWdmVE2q8J9kZbhiRtF06XVM33kO6cw2E4BoCnncQAvD_BwE)

biospecdb/apps/uploader/models.py
jamienoss: I should hope so.

FYI I didn't add this in this PR, it was already set to 128.

biospecdb/apps/uploader/admin.py
```diff
@@ -626,7 +633,7 @@ class PatientAdmin(RestrictedByCenterMixin, NestedModelAdmin):
     readonly_fields = ["created_at", "updated_at"]  # TODO: Might need specific user group.
     date_hierarchy = "updated_at"
     ordering = ("-updated_at",)
-    list_filter = ("center", "gender")
+    list_filter = ("center", "visit__observation__observable")
```
jamienoss, line 636: I also forgot to mention, that this doesn't replace the previous functionality that filters by gender type, but instead only by objects with observations of the gender observable. The former filter can be added by specifying a custom filter. See [Add-custom-admin-filter-for-gender-type](https://www.notion.so/Add-custom-admin-filter-for-gender-type-2c355dd7b5684fdcba395c498555c68a?pvs=4)

biospecdb/apps/uploader/admin.py
```diff
@@ -273,7 +276,9 @@ def __init__(self, *args, **kwargs):
         try:
             observable = next(self.observables)
             self.fields["observable"].initial = observable
-            self.fields["observable_value"].widget = self._get_widget(observable.value_class)
+            self.fields["observable_value"].widget = self._get_widget(observable.value_class,
+                                                                      choices=observable.value_choices)
+            self.fields["observable"].queryset = Observable.objects.filter(name=observable.name)
```
jamienoss, line 281: @glemson This line prevents the observable from being selectable which would then nullify the correctness of the associated widget for ``observable_value``.

Screenshot 2024-02-08 at 10 31 23 AM

biospecdb/apps/uploader/admin.py
glemson: ok-ed

biospecdb/apps/uploader/models.py
glemson: imported as python code to be run?

glemson (approved):

biospecdb/apps/uploader/models.py
jamienoss: Yep, 1000% super-sketchy, but hey 🤷 Same principle as for the QC annotators.

PR #242, opened 2024-02-09
Add ``get_column_names.py`` management command

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Add-to-README-column-names-for-bulk-upload-data-49aa99bbedeb49aba58f04ea80a11926?pvs=4)


biospecdb/apps/uploader/base_models.py
```diff
@@ -7,7 +7,27 @@
...
+                   "data",
+                   "date",
+                   "id",
+                   "spectral data file",
+                   "updated at"}
+
+        if hasattr(cls, "parse_fields_from_pandas_series"):  # Only models with this func have bulk data upload columns.
+            return {field.verbose_name.lower() for field in cls._meta.fields if not field.is_relation} - exclude
+
+        return ()
```
amitschang, line 27: Maybe better to return an empty set, to match the type returned when there is data?

biospecdb/apps/uploader/management/commands/get_column_names.py
```diff
@@ -0,0 +1,67 @@
...
+            column_names = sorted(set(column_names))
+
+            # Print column names.
+            for name in column_names:
+                self.stdout.write(name)
+
+        except CommandError:
+            raise
+        except Exception as error:
+            raise CommandError(f"An error occurred whilst collecting column names: {type(error).__name__}: {error}")
```
amitschang, line 67: this doesn't seem to add much, any particular reason to wrap in try iso just letting exception go? Is it for the benefit of some particular caller?

amitschang (approved):
Some comments, but they are minor so selecting approve

biospecdb/apps/uploader/management/commands/get_column_names.py
jamienoss: It doesn't, however, this is a command line management tool that would be better if it didn't raise. Django handles this by catching ``CommandError``.

amitschang (approved):

PR #246, opened 2024-02-14
Add health checks

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Add-healthz-9a01d55f7fce4bce8b230ba2e43e32c9?pvs=4)

Use https://github.com/revsys/django-health-check to create a ``healthz/`` health check.

TODOs:

- [ ] ~Incorporate into docker health check.~ Punt to #248 as I'm not sure how to implement this.

Screenshot 2024-02-13 at 8 03 11 PM

### DIFF
--- a/tests/test_health_check.py
+++ b/tests/test_health_check.py
@@ -1,0 +1,32 @@
+import json
+import pytest
+
+from django.test import Client
+
+
+@pytest.mark.django_db(databases=["default", "bsr"])
+class TestHealthCheck:
+    def test_healthz(self):
+        c = Client()
+        response = c.get("/healthz/")
+        assert response.status_code == 200
+
+    @pytest.mark.parametrize(("url", "header"), (("/healthz/?format=json", None),
+                                                 ("/healthz/", {"accept": "application/json"})))
+    def test_healthz_json(self, url, header):
+        c = Client()
+        response = c.get(url, headers=header)
+
+        assert response.status_code == 200
+
+        content = json.loads(response.content)
+
+        # There should be an entry for each of the installed checks:
+        assert len(content) == 4
+        assert "Cache backend: default" in content
+        assert "DatabaseBackend" in content
+        assert "DefaultFileStorageHealthCheck" in content
+        assert "MigrationsHealthCheck" in content
+
+        for x in content.values():
+            assert x.strip().lower() == "working"
--- a/tests/test_health_check.py
+++ b/tests/test_health_check.py
@@ -1,15 +1,33 @@
 import json
 import pytest
 
+from django.db import DatabaseError
 from django.test import Client
+
+from health_check.db.models import TestModel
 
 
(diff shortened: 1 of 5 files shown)


tests/test_health_check.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_healthz(self):
+        c = Client()
+        response = c.get("/healthz/")
+        assert response.status_code == 200
+
+    @pytest.mark.parametrize(("url", "header"), (("/healthz/?format=json", None),
+                                                 ("/healthz/", {"accept": "application/json"})))
+    def test_healthz_json(self, url, header):
+        c = Client()
+        response = c.get(url, headers=header)
```
amitschang, line 18: IF this supports a head request, could maybe add that as a test?

tests/test_health_check.py
```diff
@@ -0,0 +1,32 @@
...
+
+        assert response.status_code == 200
+
+        content = json.loads(response.content)
+
+        # There should be an entry for each of the installed checks:
+        assert len(content) == 4
+        assert "Cache backend: default" in content
```
amitschang, line 26: any simple way to add test that _should_ yield bad response?

amitschang (approved):
Looks good to me. Some speculative comments, but if they are not a target I'd say go ahead

tests/test_health_check.py
jamienoss: Yeah, I thought about this and can test it manually, but what I came up with was to nuke the DB file. However, for sqlite and pytest, django doesn't actually create a file but keeps the entire test DB in memory.

Any thoughts are welcome as I agree this test would be good to future sanity check.
I thought about monkey patching but couldn't actually locate the DB namespace - but perhaps I didn't try hard enough.

tests/test_health_check.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.mark.django_db(databases=["default", "bsr"])
+class TestHealthCheck:
+    @pytest.fixture
+    def sabotage_db(self, monkeypatch):
+        def save(*args, **kwargs):
+            raise DatabaseError("Nice try")
+
+        # Monkeypatch health_check's TestModel to raise when testing a DB write.
+        monkeypatch.setattr(TestModel, "save", save)
```
jamienoss, line 18: @amitschang I tried messing with the cache but monkeypatching wasn't working since it was already too late.

Looking at https://github.com/revsys/django-health-check/blob/master/health_check/db/backends.py#L14
it was a lot easier to just mess with the DB check.

tests/test_health_check.py
jamienoss: See https://github.com/ssec-jhu/biospecdb/pull/246#discussion_r1490117270

amitschang (approved):
Looks great, thanks!

PR #247, opened 2024-02-14
Return 301 (permanent), not 307 (temporary) redirect http resp

jamienoss:


jamienoss:
FYI I'll admin merge this once tests pass.

PR #248, opened 2024-02-14
Docker healthchecks

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Add-docker-healthcheck-e667f1aec7b74f8399dc4e01c639677c?pvs=4)
Relates to #246

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM python:3.11
 
+RUN ln -sf /bin/bash /bin/sh
+
 ENV PYTHONUNBUFFERED=1
 
 WORKDIR /app
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,6 +62,12 @@ services:
       && mkdir -p run
       && SECRET_KEY=$(cat /run/secrets/django_secret_key) python manage.py collectstatic --clear --noinput
       && SECRET_KEY=$(cat /run/secrets/django_secret_key) gunicorn -c config/gunicorn/prd.py"
+    healthcheck:
+      test: ["CMD-SHELL", "curl --head --fail --header 'X-Forwarded-Proto: https' http://localhost:8000/healthz/ || exit 1"]
+      interval: 60s
+      timeout: 10s
+      retries: 3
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -62,6 +62,12 @@ services:
       && mkdir -p run
       && SECRET_KEY=$(cat /run/secrets/django_secret_key) python manage.py collectstatic --clear --noinput
       && SECRET_KEY=$(cat /run/secrets/django_secret_key) gunicorn -c config/gunicorn/prd.py"
+    healthcheck:
+      test: curl --fail http://localhost:8000/healthz/?format=json || exit 1
```
jamienoss, line 66: I'm not quite sure how to do this given that Django is set up to redirect all to https, thus this cmd always gives a 301.

Using ``curl --fail https://localhost/healthz/?format=json || exit 1`` also won't work with the ``nginx::web::condition::service_healthy`` since nginx _is_ the SSL layer.

jamienoss: Could easily add a separate gunicorn healthz endpoint to be used for ``nginx::web::condition::service_healthy`` 🤷 .

jamienoss: As @amitschang suggested, adding the ``X-Forwarded-Proto`` header resolved this issue.

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM python:3.11
 
+RUN ln -sf /bin/bash /bin/sh
```
jamienoss, line 3: Added so the terminal automatically used ``bash`` over ``sh`` such that arrow keys then worked.

docker-compose.yml
```diff
@@ -82,7 +88,8 @@ services:
       - bulk-upload-files:/raw_data
       - dataset-catalog-files:/datasets
     depends_on:
-      - web
+      web:
+        condition: service_healthy
```
jamienoss, line 92: At present, this is the only utilized benefit of the health check.

amitschang (approved):
Looks good to me

PR #255, opened 2024-02-28
AWS AppRunner setup w/postgres RDS

jamienoss:
Resolves [Notion ticket](https://www.notion.so/AWS-App-runner-deployment-and-postgres-usage-e47c7edd0cd04854846356f9f82e0007?pvs=4)

- [x] Depends on #265 
- [x] Depends on #266

## Changes:
- [x] Read only DB access for sql-explorer app. [psql docs](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-DEFAULT-TRANSACTION-READ-ONLY). [Resolves Notion ticket](https://www.notion.so/Make-SQL-explorer-read-only-b186c89682754758a09f4854bbc46baa?pvs=4).
- [x] Add logging (resolves [Notion ticket](https://www.notion.so/Add-logging-e955be3f79784fe5858e95013eb681da?pvs=4))
- [x] Add ``apprunner.yml``, ``biospecdb/settings/aws.py``, and ``scripts/prd/deploy.sh`` for AWs app runner config.
- [x] Add ``makesuperuser`` Django management command to wrap ``createsuperuser`` but not fail on existing user creation.
- [x] Serve static files from gunicorn using ``whitenoise`` package.
- [x] Add deployment script.
- [x] Add django settings file for AWS deployment.
- [x] Postgres
  - [x] Switch DB from sqlite to postgres.
  - [x] Fix sql view creation for postgres.
    - [x] Drop view even for postgres since column names can't be updated, only appended. This won't work if observables are deleted. See [docs](https://www.postgresql.org/docs/current/sql-createview.html):
        > CREATE OR REPLACE VIEW is similar, but if a view of the same name already exists, it is replaced. The new query must generate the same columns that were generated by the existing view query (that is, the same column names in the same order and with the same data types), but it may add additional columns to the end of the list. The calculations giving rise to the output columns may be completely different.
    - [x] Refactor management command ``update_sql_views`` to not update all views. Due to psql's ``CASCADE`` view deletion, automatic view deletion requires that they get deleted and created in a specific order. To do this, the cmd would have to correctly graph the views and deps to correctly create this order - which just isn't worth it. Without this correct order, views created can be subsequently dropped and left non-existent if sub-views are created (and thus dropped) afterward.
    - [ ] ~[MATERIALIZED](https://www.postgresql.org/docs/current/rules-materializedviews.html) vs normal views in postgres and vs those with sqlite?~ Simpler not to due to creation testing issues and errors thrown destroying transactions.
  - [x] Catch ``ProgrammingError`` when catching``OpertaionalError``.
  - [x] Don't store app version as model field default as it causes migration issues with postgres (and isn't a good idea anyhow), instead use a callable.
  - [x] Fix issue with ``makemigrations`` in relation to observable access during ``ObservationInline.__init__``. 
- [x] Don't serve static files from nginx for ``docker-compose.yml``.
- [x] S3
  - [x] Store and serve media files from S3, not local fs and nginx.
  - [x] Encryption at rest. S3's default see [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-encryption.html).
  - [x] Check file cleanup commands still work. Moved to #265 
  - [x] Add cron job to ``deploy.sh``. Depends on #266
  - [x] ~``django-storage`` docs imply that files actually get deleted when DB entries are, check this and it would solve the above cleanup issue.~ Maybe, but doesn't matter as DB deletions happen from the admin forms using bulk deletions which are raw SQL and side step a model's ``delete`` method. Moved to #265 
  - [x] Check exporters work with S3, e.g., zipping all data files together.
  - [x] Check bug in sqlexplorer app for full_patient. Looks like data isn't being loaded, resulting in an exception and charts not being loaded. ~Could just be an issue with the data as we're using the test data.~ This probably more like an S3 vs fs issue.


apprunner.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - pip3 install pipenv
+      - pipenv install -r requirements/prd.txt
+  env:
+    - name: DJANGO_SETTINGS_MODULE
+      value: "biospecdb.settings.prd"
+run:
+  runtime-version: 3.11
+  pre-run:
+    - pip3 install pipenv
+    - pipenv install -r requirements/prd.txt
```
jamienoss, line 15: Doing this twice is required for python 3.11 as it uses the "revised app runner", see [Python 3.11 (revised App Runner build)](https://docs.aws.amazon.com/apprunner/latest/dg/service-source-code-python.html).

biospecdb/apps/catalog/migrations/0001_initial.py
```diff
@@ -12,7 +12,7 @@ class Migration(migrations.Migration):
     initial = True
 
     dependencies = [
-        ('explorer', '0012_alter_queryfavorite_query_alter_queryfavorite_user'),
+        ('explorer', '0013_querylog_error_querylog_success'),
```
jamienoss, line 15: This should have been updated in #260

biospecdb/apps/uploader/admin.py
```diff
@@ -5,7 +5,7 @@
 from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.db.models import Manager, Q
-from django.db.utils import OperationalError
+from django.db.utils import OperationalError, ProgrammingError
```
jamienoss, line 8: postgresql django API throws a ``ProgrammingError`` whilst that for sqlite throws an ``OperationalError``.

biospecdb/apps/uploader/base_models.py
```diff
@@ -102,17 +102,24 @@ def update_view(cls, *args, **kwargs):
         if "db" not in kwargs:
             kwargs["db"] = cls.db
 
+        sql, params = cls.sql()
+
+        if connection.vendor == "postgresql":
+            # Drop views on the way down so that all are gone before any are created.
```
jamienoss, line 108: FYI This would also work for sqlite.

config/gunicorn/prd.py
```diff
@@ -33,14 +35,14 @@
...
 
 # Write access and error info to:
-accesslog = "log/prd_access.log"
-errorlog = "log/prd_error.log"
+# accesslog = "log/prd_access.log"
+# errorlog = "log/prd_error.log"
 
 # Redirect stdout/stderr to log file
-capture_output = True
+capture_output = False
```
jamienoss, line 42: Don't capture such that AWS apprunner can catch logs.

nginx/nginx.conf
```diff
@@ -36,10 +36,6 @@ server {
         client_max_body_size    100M;
     }
 
-    location /static/ {
```
jamienoss, line 39: Now served using the ``whitenoise`` app.

jamienoss:
@amitschang Sorry for the lengthy PR 😬 . I really should have split this out into a few PRs but here we are. Cheers.

biospecdb/apps/uploader/models.py
```diff
@@ -874,7 +874,7 @@ def sql(cls):
             else:
                 value = "observable_value"
             d.append(f"max(case when observable = '{observable.name}' then {value} else null end) as "
-                     f"[{observable.name}]")
+                     f"{observable.name}")
```
amitschang, line 877: suspicious about the removed quoting, is it guaranteed that the name be OK without quoting? Is `[` a problem in postgres?

biospecdb/apps/uploader/sql.py
```diff
@@ -14,35 +14,56 @@ def execute_sql(sql, db=None, params=None):
...
-    except OperationalError as error:
-        if "no such view:" not in str(error):
-            raise
+    if connection.vendor == "postgresql":
+        execute_sql(f"drop view if exists {view} cascade", db=db)  # nosec B608
+    elif connection.vendor == "sqlite":
+        execute_sql(f"drop view if exists {view}", db=db)  # nosec B608
+    else:
+        # Note: This may not be necessarily true, but test view correctness against vendor and explicitly add to the
+        # above.
```
amitschang, line 38: :+1:

biospecdb/settings/base.py
```diff
@@ -79,21 +81,84 @@
...
+            "USER": os.getenv("DB_BSR_USER"),
+            "PASSWORD": os.getenv("DB_BSR_PASSWORD"),
+        },
+        "bsr_readonly": {
+            "ENGINE": "django.db.backends.postgresql",
+            "NAME": "bsr",
+            "HOST": os.getenv("DB_BSR_HOST"),
+            "PORT": os.getenv("DB_BSR_PORT"),
+            "USER": os.getenv("DB_BSR_USER"),
+            "PASSWORD": os.getenv("DB_BSR_PASSWORD"),
```
amitschang, line 153: should readonly user/pass be same as readwrite?

jamienoss:
@amitschang FYI I just added another commit 6087f702d3b59b39dd36e399da500abf7f451024.

biospecdb/apps/uploader/management/commands/update_sql_views.py
```diff
@@ -1,23 +1,24 @@
...
-from uploader.base_models import SqlView
-import uploader.models
-
 
 class Command(BaseCommand):
     help = "Create/update all custom SQL views in uploader.models."
 
+    def add_arguments(self, parser):
+        parser.add_argument("view",
+                            default=None,
```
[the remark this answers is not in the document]
jamienoss: Good catch! Thanks

biospecdb/apps/uploader/models.py
jamienoss: `[` _is_ a problem with postgres, and technically should be too for sqlite, because it's not valid SQL. However, I guess sqlite is more forgiving. It's specific to sql server (MS).

This has been a hidden bug from day 1.

amitschang: haha, didn't realize cause we are such a mssql shop, but still ought to be quoted appropriately, no?

biospecdb/apps/uploader/management/commands/update_sql_views.py
```diff
@@ -1,23 +1,24 @@
...
+            for model in apps.get_models():
+                if hasattr(model, "_meta") and hasattr(model._meta, "db_table") and model._meta.db_table == view:
+                    self.stdout.write(f"Creating/updating SQL view: {view}...", ending='')
+                    model.update_view()
                     self.stdout.write(
                         self.style.SUCCESS("[Done]")
                     )
-                except Exception as error:
-                    raise CommandError(f"An error occurred whilst trying to create/update the sql-view: '{name}' -"
-                                       f" '{error}'")
```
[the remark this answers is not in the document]
jamienoss: Nope, another good catch, thanks!

biospecdb/apps/uploader/models.py
jamienoss: It does and is, see lines above and below.

jamienoss: 🤔 Good point

jamienoss: Trying it out, it complains of a syntax error so I guess not 🤷 
```python
 CommandError: An error occurred whilst trying to create/update the sql-view: 'full_patient' - 'syntax error at or near "'Ct_gene_N'"
LINE 4: ... cast(observable_value AS REAL) else null end) as 'Ct_gene_N...
                                                             ^'
```
I'll defer this to @glemson since this is their creation and my SQL obviously sucks lol.

jamienoss: I think, because, by that point in the stmnt, the name is an alias it shouldn't have them.

biospecdb/apps/uploader/sql.py
```diff
@@ -14,35 +14,56 @@ def execute_sql(sql, db=None, params=None):
...
+        try:
+            if result := cursor.fetchall():
+                columns = [col[0] for col in cursor.description]
+                return [dict(zip(columns, row)) for row in result]
+        except ProgrammingError as error:
+            # With postgres, fetchall() raises rather than returning None (seems like a bug).
+            if "the last operation didn't produce a result" in str(error) or "no result available" in str(error):
+                pass
+            else:
+                raise
```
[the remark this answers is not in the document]
jamienoss: This doesn't work.
Also seems like a bug in ``pyscopg`` https://github.com/psycopg/psycopg/blob/558cbe7ef7ef6a6d1f6c65e368aaaf9f9a10108e/psycopg/psycopg/_cursor_base.py#L585 is raising our error when ``status==1``. Looking at https://github.com/psycopg/psycopg/blob/558cbe7ef7ef6a6d1f6c65e368aaaf9f9a10108e/psycopg/psycopg/pq/_enums.py#L67 ``1 := COMMAND_OK``, seems like a bug considering that ``fetchall()``should return an empty iter if there's nothing to consume. 🤷 
I'm rolling this back and punting to [Notion ticket](https://www.notion.so/Handle-pscopg-cursor-rowcount-cursor-fetchall-bug-8c09f0e181c643d3bde6d7f7671083f3?pvs=4).

biospecdb/settings/base.py
```diff
@@ -79,21 +81,84 @@
...
+        },
+        "bsr_readonly": {
+            "ENGINE": "django.db.backends.postgresql",
+            "NAME": "bsr",
+            "HOST": os.getenv("DB_BSR_HOST"),
+            "PORT": os.getenv("DB_BSR_PORT"),
+            "USER": os.getenv("DB_BSR_USER_READONLY"),
+            "PASSWORD": os.getenv("DB_BSR_PASSWORD_READONLY"),
```
jamienoss, line 153: @amitschang I'm also gonna roll this back and punt to [Notion ticket](https://www.notion.so/Setup-readonly-psql-user-a3b45f2fee9444c8befc9eed2740b0c9?pvs=4) as I can't get it working.

biospecdb/settings/base.py
jamienoss: See  https://github.com/ssec-jhu/biospecdb/pull/255/files#r1522066389

amitschang (approved):
Looks good to go

PR #262, opened 2024-03-04
Add version view

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Add-path-to-return-app-version-81344ab64d28469e82d6a03bcd8f5f81?pvs=4)

``_version__`` is populated from ``_version`` which is created by the build - this may need some alteration with respect to #255 (which will take place in that PR).


amitschang (approved):
Looks good to me

jamienoss:
This doesn't quite work on aws apprunner. I'm going to merge and then take this up over at https://github.com/RISPaDD/biospecdb

PR #263, opened 2024-03-04
Separate out deps into more req files

jamienoss:
Copy of https://github.com/ssec-jhu/base-template/pull/113


amitschang (approved):
super!

PR #220, opened 2024-01-26
Add prune_files management command to delete orphaned files

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Django-doesn-t-delete-files-on-disk-when-dereferenced-from-the-DB-e12623af4c8d471db8a39b6dceb3cfdb?pvs=4)
Resolves #39 

## TODO:
- [x] Add tests
- [ ] Add cronjob to docker? PUNT - perhaps even indefinitely as it may be better to manually do this.

PR #265, opened 2024-03-11
Use Django storage API for all file access

jamienoss:
Resolves [Notion ticket](https://www.notion.so/Use-Django-storage-API-357f4fa8bbf1469b818561f88b78e3aa?pvs=4)

Django's storage API must be used, rather than Python's functionality, i.e., ``open``, ``os.remove``, so that non-fs storages, i.e., AWS S3, can be utilized.

Reimplements ``prun_files`` from #220

Required by #255

I've manually tested that this works with S3.
 - [x] Media files written to S3.
 - [x] Media files downloadable from S3, via BSR web app.
 - [x] Data archived/zipped correctly for:
   - [x] Cataloged datasets.
   - [x] From SQL explorer.

### DIFF
--- a/biospecdb/apps/uploader/exporters.py
+++ b/biospecdb/apps/uploader/exporters.py
@@ -3,7 +3,6 @@ import tempfile
 import zipfile
 
 from django.conf import settings
-from django.core.files.storage import storages
 from django.utils.module_loading import import_string
 import explorer.exporters
 import pandas as pd
@@ -74,6 +73,7 @@ class ZipSpectralDataMixin:
 
         data_files = []
         if include_data_files:
+            storage = SpectralData.data.field.storage
             # Collect SpectralData files and zip along with query data from self._get_output().
             if settings.EXPLORER_DATA_EXPORTERS_ALLOW_DATA_FILE_ALIAS:
                 # Spectral data files are modeled by the Spectraldata.data field, however, the sql query could have
@@ -109,11 +109,11 @@ class ZipSpectralDataMixin:
                 # Add all data files to zipfile.
                 for filename in data_files:
                     try:
-                        archive.write(storages["default"].path(filename), arcname=filename)
+                        archive.write(storage.path(filename), arcname=filename)
                     except NotImplementedError:
-                        # storages["default"].path() will raise NotImplementedError for remote storages like S3. In this
+                        # storage.path() will raise NotImplementedError for remote storages like S3. In this
                         # scenario, open and read all the file contents to zip.
-                        with storages["default"].open(filename) as fp:
+                        with storage.open(filename) as fp:
                             data = fp.read()
                         archive.writestr(filename, data)
             temp.seek(0)
--- a/biospecdb/apps/uploader/exporters.py
+++ b/biospecdb/apps/uploader/exporters.py
@@ -1,9 +1,9 @@
 import hashlib
-from pathlib import Path
 import tempfile
 import zipfile
 
 from django.conf import settings
+from django.core.files.storage import storages
 from django.utils.module_loading import import_string
 import explorer.exporters
 import pandas as pd
@@ -108,7 +108,14 @@ class ZipSpectralDataMixin:
 
                 # Add all data files to zipfile.
                 for filename in data_files:
-                    archive.write(Path(filename))
+                    try:
+                        archive.write(storages["default"].path(filename), arcname=filename)
+                    except NotImplementedError:
+                        # storages["default"].path() will raise NotImplementedError for remote storages like S3. In this
+                        # scenario, open and read all the file contents to zip.
+                        with storages["default"].open(filename) as fp:
+                            data = fp.read()
+                        archive.writestr(filename, data)
             temp.seek(0)
             output = temp
             self.is_zip = True
--- a/biospecdb/apps/uploader/management/commands/prune_files.py
+++ b/biospecdb/apps/uploader/management/commands/prune_files.py
@@ -24,23 +24,21 @@ class Command(BaseCommand):
                 self.stdout.write(self.style.WARNING("No orphaned files detected."))
                 return
 
-            n_orphaned_files = 0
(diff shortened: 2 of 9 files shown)


biospecdb/apps/uploader/tests/test_management_commands.py
```diff
@@ -17,21 +17,41 @@ def test_empty(self):
...
+        assert get_file_count(UploadedFile.UPLOAD_DIR) == 2
+        assert get_file_count(SpectralData.UPLOAD_DIR) == 10
+
+        # Delete objects and files (if delete).
+        def delete_objs(model):
+            for obj in model.objects.all():
+                obj.delete(delete_files=delete)
+            assert model.objects.count() == 0
+        delete_objs(UploadedFile)
+        delete_objs(SpectralData)
```
jamienoss, line 43: TBH, I'm not really sure how this test worked before without this, whoops.

jamienoss:
@amitschang demoted to draft, csv exporters aren't working properly...

biospecdb/apps/uploader/exporters.py
```diff
@@ -108,7 +108,14 @@ def get_file_output(self,
...
                 for filename in data_files:
-                    archive.write(Path(filename))
+                    try:
+                        archive.write(storages["default"].path(filename), arcname=filename)
+                    except NotImplementedError:
+                        # storages["default"].path() will raise NotImplementedError for remote storages like S3. In this
+                        # scenario, open and read all the file contents to zip.
+                        with storages["default"].open(filename) as fp:
+                            data = fp.read()
+                        archive.writestr(filename, data)
```
jamienoss, line 118: @amitschang Disregard https://github.com/ssec-jhu/biospecdb/pull/265#issuecomment-1988380881, must have been user error or some quasi-data state between git branches. 

The zipping worked mostly as expected, however, the spectral data files were getting added with their full parent dir list including the repo dir on up.

I've retested for both local FS and S3 and this isn't happening, weird. 🤷

biospecdb/apps/uploader/management/commands/prune_files.py
```diff
@@ -16,38 +12,42 @@ def add_arguments(self, parser):
...
+                    if orphaned_files:
+                        all_orphaned_files.append((storage, orphaned_files))
+
+            if not all_orphaned_files:
                 self.stdout.write(self.style.WARNING("No orphaned files detected."))
                 return
 
+            n_orphaned_files = 0
+            for x in all_orphaned_files:
+                n_orphaned_files += len(x[1])
```
amitschang, line 29: or

```py
n_orphaned_files = sum(len(x) for _,x in all_orphaned_files)
```

biospecdb/apps/uploader/management/commands/prune_files.py
```diff
@@ -16,38 +12,42 @@ def add_arguments(self, parser):
...
-                else:
-                    self.stdout.write(f"Deleting {msg}")
-                    try:
-                        os.remove(file)
-                    except FileNotFoundError:
-                        continue
+            n_deleted = 0
+            for storage, files in all_orphaned_files:
+                for file in files:
+                    n_deleted += 1
```
amitschang, line 35: This is not quite true, and wouldn't in principal differ from `n_orphaned_files` (except in the case where outer exception, then it would also not necessarily be correct) - should it rather go after `storage.delete(file)`?

biospecdb/apps/uploader/exporters.py
```diff
@@ -108,7 +108,14 @@ def get_file_output(self,
 
                 # Add all data files to zipfile.
                 for filename in data_files:
-                    archive.write(Path(filename))
+                    try:
+                        archive.write(storages["default"].path(filename), arcname=filename)
```
amitschang, line 112: I see storage passed around as argument (e.g. in `get_orphaned_files`), implying they could be different - just want to be sure `"default"` is intentional and always correct as called in the diff. I don't know enough to tell

biospecdb/apps/uploader/models.py
```diff
@@ -134,16 +133,32 @@ def save(self, *args, **kwargs):
...
     def adelete(self, *args, **kwargs):
         raise NotImplementedError
 
+    @classmethod
+    def get_orphan_files(cls):
+        storage = cls.meta_data_file.field.storage
+        path = Path(settings.MEDIA_ROOT) / cls.UPLOAD_DIR
+        # Collect all stored media files.
+        try:
+            fs_files = set([str(path / x) for x in storage.listdir(str(path))[1]])
```
amitschang, line 153: this differs just slightly from the other orphan files method (`listdir(str(path))` vs `listdir(path)`), is this intentional? They (the `get_orphan_files` methods) also look similar enough that maybe a util function to do the main work with different data would be useful?

amitschang (commented):
Some comments/questions

biospecdb/apps/uploader/models.py
jamienoss: >this differs just slightly from the other orphan files method (listdir(str(path)) vs listdir(path)), is this intentional? 

Yes, because the filenames returned from the DB query inc. the dir, but since you pass in the dir to ``listdir`` the names don't inc. it - for comparison this is needed.

>They (the get_orphan_files methods) also look similar enough that maybe a util function to do the main work with different data would be useful?

I originally thought about this, and even wasted time with it, however, there's no convenient way to factor this since multiple ``FileFields`` can use the same storage dir, as is the case for bulk uploaded meta and spectral data files. To account for this uniqueness, I thought it better to go in the model that way code can be as custom and as explicit as it need be.

biospecdb/apps/uploader/exporters.py
jamienoss: This should actually be ``SpectralData.data.field.storage``.

amitschang (approved):
cool, thanks for changes and comments

PR #290, opened 2024-03-15
Add utility funcs & scripts to retrieve AWS secrets, parsed from apprunner.yaml

jamienoss:


amitschang (approved):
Looks fine, but I do agree that it would be more ideal if the apprunner yaml could be the source for the unique ids.

amitschang (dismissed):
Looks better!

jamienoss:
@amitschang thanks for the great suggestion and help!

amitschang (approved):

PR #297, opened 2024-03-16
295 Fix center deletion

### LINKED ISSUE #295: Center not being correctly deleted across both DBs

jamienoss:
Resolves #295

So this isn't complete, whilst this will now delete all centers upon request, related/protected items will not necessarily be correctly displayed in the to_delete list. They will, however, be correctly protected, so this PR is safe in that regard. 

Back story: ``Center`` is a duplicated table/model existing on both ``admin`` (default) and ``bsr`` databases (we needed the DB split for the separation of user data from actual data). Relations can't exist across DBs even for PostgreSQL.

One stop-gap measure is to swap the admins, i.e., ``user.models.Center`` is related to only the ``user.models.User``, whilst ``uploader.models.Center`` is related to pretty much all of the data objects via one relation or another. It would be better to display all the data that would need to be deleted compared to just a few users.
The next stop-gap measure would be to add both admins.

For this PR, I'll go with 2nd stop-gap and display both admins for ``Center`` - these are only visible to admin users anyhow.

The correct fix is to automatically do both and list all relations across both DBs. Punted to #303 as I'll get around to this shortly but it would still be better to merge in the above.

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -14,6 +14,7 @@ from biospecdb.util import to_bool
 from .models import BioSample, Observable, Instrument, Patient, SpectralData, Observation, UploadedFile, Visit,\
     QCAnnotator, QCAnnotation, Center, get_center, BioSampleType, SpectraMeasurementType
 from uploader.forms import ModelForm
+from user.admin import CenterAdmin as UserCenterAdmin
 
 User = get_user_model()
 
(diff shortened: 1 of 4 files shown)


biospecdb/apps/user/migrations/0001_initial.py
```diff
@@ -46,7 +46,7 @@ class Migration(migrations.Migration):
                 ('is_sqluser_view', models.BooleanField(default=False, help_text='Designates whether the user can log into the SQL explorer app with permissions to only view and execute existing queries.', verbose_name='SQL explorer user status (view/execute existing queries only)')),
                 ('is_sqluser_change', models.BooleanField(default=False, help_text='Designates whether the user can log into the SQL explorer app with permissions to view/add/change/delete/execute queries.', verbose_name='SQL explorer user status (view/add/change/delete/execute)')),
                 ('is_catalogviewer', models.BooleanField(default=False, help_text='Designates whether the user can log into the Dataset Catalog app. (readonly)', verbose_name='Dataset Catalog user status (readonly)')),
-                ('center', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='user', to='user.center')),
+                ('center', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='user', to='user.center')),
```
jamienoss, line 49: This should be discussed.

PR #293, opened 2024-03-15
292 Add extended sql explorer perms

jamienoss:
Resolves #292

@Alvaro-FG FYI I've added the additional sql-explorer permission flag. So now there's
 * ``is_sqluser_view``
 * ``is_sqluser_change``
 Also, as a reminder, use these sparingly as they will make the entire DB visible to the user.

PR #299, opened 2024-03-16
Fix migrations from #293

jamienoss:
With a live DB we can't nuke the initial migrations as we have been in dev, they now need to actually be migrated from migration to migration.


scripts/dev/rebuild_postgres.sh
```diff
@@ -69,4 +64,4 @@ python manage.py prune_files
 
 # Creat superuser.
 # Note: This center ID is that for the SSEC and the default password is "admin".
-DJANGO_SUPERUSER_PASSWORD="${DJANGO_SUPERUSER_PASSWORD:-admin}" python manage.py createsuperuser --noinput --username=admin --email=admin@jhu.edu --center=d61f1c2a-9c0a-4309-a031-ab5b8d2106b0
```
jamienoss, line 72: Missing from an older PR.

PR #327, opened 2024-04-17
322 Fix broken inline observations

### LINKED ISSUE #322: Editing patients asks for comorbidities
With the mock patients, when I try to edit one of them and add, for example, a test name, when trying to save it it requires to add comorbidities

jamienoss:
Resolves #322

### DIFF
--- a/biospecdb/apps/uploader/admin.py
+++ b/biospecdb/apps/uploader/admin.py
@@ -357,14 +357,21 @@ class ObservationInlineForm(ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        try:
-            observable = next(self.observables)
-            self.fields["observable"].initial = observable
+
+        observable = None
+        if instance := kwargs.get("instance"):
+            observable = instance.observable
+        else:
+            try:
+                observable = next(self.observables)
+                self.fields["observable"].initial = observable
+            except StopIteration:
+                pass
+
+        if observable:
+            self.fields["observable"].queryset = Observable.objects.filter(name=observable.name)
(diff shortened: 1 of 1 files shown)


biospecdb/apps/uploader/admin.py
```diff
@@ -357,14 +357,21 @@ def _get_widget(value_class, choices=()):
...
+            observable = instance.observable
+        else:
+            try:
+                observable = next(self.observables)
+                self.fields["observable"].initial = observable
+            except StopIteration:
+                pass
+
+        if observable:
+            self.fields["observable"].queryset = Observable.objects.filter(name=observable.name)
```
jamienoss, line 372: There were three bugs altogether, the main one being that setting ``queryset`` conflicted with ``initial``, which meant that observables weren't being correctly chosen, which blocked when saving an edit - since the observables must be specified since the values were for existing observations.

biospecdb/apps/uploader/admin.py
```diff
@@ -408,16 +415,19 @@ def get_extra(self, request, obj=None, **kwargs):
         """ Only list extra inline forms when no data exists, i.e., new patient form. """
         if obj and obj.pk and obj.observation.count():
```
jamienoss, line 416: For some reason, ``obj`` was no longer being passed in resulting in an incorrect number of "extra" inline forms being populated.

jamienoss:
The last bug, was very pre-existing, in that only "global-center" inline observables were listed for existing patient observations, even when non-global observations may have existed. This was coincidentally fixed by using ``observable = kwargs["instance"]`` in the inline form init.

PR #333, opened 2024-04-19
Fail early in model.full_clean so field validators bubble up

### LINKED ISSUE #332: Server Error 500
See [instance/b8bddae432d44606bb4926b3d4cbdc60]

2024-04-18T22:54:24.871+01:00

I believe the issue is when the file uploaded is not of certain formats it does not provide an alert but rather a Server Error 500

jamienoss:
Resolves #332 

My understanding from the Django docs is that ``ModelForm`` should pull the validators from the model fields such that when the form is validated, the field validators are called. This does not seem to happen.

#250 approached this before, however, it still relied on the field validators running.

Changes:
 * Extend #250 to all model forms in ``uploader.admin`` (doesn't bother inline admins and the ``user`` and ``catalog`` apps 🤷 ).
 * Hard copy ``django.db.models.Model.full_clean`` and add ``fail_early`` option such that field validations are raised before ``Model.clean`` is called.


biospecdb/apps/uploader/admin.py
```diff
@@ -19,6 +19,13 @@
 User = get_user_model()
 
 
+class ModelAdmin(admin.ModelAdmin):
+    def get_form(self, request, obj=None, **kwargs):
+        form = super().get_form(request, obj=obj, **kwargs)
+        setattr(form, "_post_clean", ModelForm._post_clean)
+        return form
```
jamienoss, line 26: This code extends #250 to all uploader admins (that aren't inlines)

biospecdb/apps/uploader/base_models.py
```diff
@@ -33,6 +35,66 @@ def get_column_names(cls, help_text=False):
...
+            exclude = set()
+        else:
+            exclude = set(exclude)
+
+        try:
+            self.clean_fields(exclude=exclude)
+        except ValidationError as e:
+            errors = e.update_error_dict(errors)
+            if fail_early:
+                raise ValidationError(errors)
```
jamienoss, line 60: This (and the repeats below) are the only code added to this func.

biospecdb/apps/uploader/models.py
```diff
@@ -123,19 +123,24 @@ def join_with_validation(meta_data, spec_data):
...
-        spec_data = uploader.io.read_spectral_data_table(self.spectral_data_file.file,
-                                                         index_column=settings.BULK_UPLOAD_INDEX_COLUMN_NAME)
-        # Validate.
-        UploadedFile.validate_lengths(meta_data, spec_data)
-        # This uses a join so returns the joined data so that it doesn't go to waste if needed, which it is here.
-        joined_data = UploadedFile.join_with_validation(meta_data, spec_data)
-
-        # Ingest into DB.
-        save_data_to_db(None, None, center=self.center, joined_data=joined_data, dry_run=dry_run)
+        try:
```
jamienoss, line 126: Wrap any exception and raise as a generic and non-descript validation error.

PR #337, opened 2024-05-01
Add user documentation

jamienoss:
The new docs can be viewed [here](https://biospecdb--337.org.readthedocs.build/en/337/usage.html). Note: some links will not work as expected as they are relative links intended to take you to the host site itself.

Changes:

- This adds a user manual for instructions on how to use the website (from a user perspective).
- When deployed this will host the docs from the live website - it will be completely public though.
- Adds link to docs from landing page.
- Adds link to password reset form from landing page.
- Adds more docstrings.
- Improves API sphinx docs.

Note: I intentionally omitted docs from an admin perspective so as not to expose potential issues.


docs/_static/.DS_Store
amitschang, line 1: did you mean to check this in?

docs/_static/confirm_delete.png
amitschang, line 1: possible to get a shorter image that contains the relevant button?

docs/source/data_collection.rst
```diff
@@ -0,0 +1,192 @@
...
+^^^^^
+
+The central `data collection page <../../data/>`_  looks like the following.
+
+.. image:: ../_static/data_collection.png
+
+From the central `data collection page <../../data/>`_ you can view, add, and edit (permissions required) the following types of data forms.
+Each form below includes all forms for those listed below it, excluding that for Bulk Data Uploads.
+
+ - Patients_: This is the principle form for collecting data for new (or returning) individual patients. **This is probably the only form you need to use.**
```
amitschang, line 14: Might be helpful to elaborate a bit on why this is probably the only form needed

docs/source/data_collection.rst
```diff
@@ -0,0 +1,192 @@
...
+Patients
+^^^^^^^^
+
+Below is an example of the fields present for a patient form.
+The ``Patient ID`` is an auto generated `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_, however, the ``Patient CID`` is provided by the user adding the patient.
+The ``center`` field will automatically be populated with your user center. This is neither optional nor editable.
+
+.. image:: ../_static/patient_fields.png
+
+.. warning:: The ``Patient CID`` is optional, however, if provided it must **NOT** contain PII. It **MUST** be either a *fully anonymized* or a *pseudonymized* identifier.
```
amitschang, line 89: if available, might be helpful to link to a resource on these terms (fully/pseudo-anonymized)

docs/source/datasets.rst
```diff
@@ -0,0 +1,71 @@
...
+Clicking on the dataset ``NAME`` will then take you to the dataset details.
+
+.. image:: ../_static/dataset_view.png
+
+Details are organized and some will be collapsed, when expanded expect something similar to the following.
+
+.. image:: ../_static/dataset_view_expanded.png
+
+
+Fields
```
amitschang, line 36: don't the fields have description in code? If so, might be simpler to refer to the form or relevant page so this doesn't have to be maintained if/when fields added.

amitschang (commented):
A few inline comments - one overall comment is that some images are quite long, could we find abbreviated versions? Or perhaps there is a way to "ellipsize" them?

docs/_static/.DS_Store
jamienoss: nope, thanks

docs/source/data_collection.rst
jamienoss: Yeah, I thought about that but didn't want to go there (more than we have in this very stmnt). It's not our place to advise on these topics in something as formal and outward-facing as this doc.

docs/_static/.DS_Store
jamienoss: ...and I didn't, at least not in this PR. It's been in there since the initial commit.

docs/_static/confirm_delete.png
jamienoss: The emphasis is not just the delete but to give an example of how much they could expect to see because of the cascade.

docs/_static/.DS_Store
jamienoss: https://github.com/ssec-jhu/opsins-discovery/pull/109

jamienoss: https://github.com/ssec-jhu/base-template/pull/132

jamienoss:
>one overall comment is that some images are quite long, could we find abbreviated versions? Or perhaps there is a way to "ellipsize" them?

I think the only two that are long are the deletion page and the full patient form, both of which I'd like to keep as such to explicitly show their length and content.

docs/source/datasets.rst
jamienoss: No, they didn't but do now.

However, not all of these are class attributes. I thought about making some properties but decided against it. Either way, I think it's better to still keep this here explicitly (even if not dry) unless you know of a way to directly insert the field text here and not just link out to the API docs?

docs/source/data_collection.rst
amitschang: Perhaps this should be left out altoghether, then. It is likely the case that nowhere in the system should have PII in any case - but depending on where running this is up to the admin to sort out

jamienoss: Could be, but this is at least true for both GDPR and HIPAA so relevant.

Is there an issue with it being in there?

jamienoss: @Alvaro-FG what's your preference?

amitschang (approved):

PR #349, opened 2024-05-14
Protect media files

jamienoss:
The following is true for local deployment. AWS deployment uses django-storages for serving data files from S3 and security around these still needs to be determined.

Spectral data files were only very weakly protected by their long UUID filename and URL path. However, links where still downloadable post-logout which is a huge security issue. Additionally, no protection existed against users sharing URL links with others.

Whilst the spectral data files, were weakly protected as described above, the bulk upload files were not protected at all!

### DIFF
--- a/biospecdb/urls.py
+++ b/biospecdb/urls.py
@@ -17,11 +17,12 @@ Including another URLconf
 from decorator_include import decorator_include
 from django.contrib import admin
 from django.contrib.auth import views as auth_views
-from django.contrib.auth.decorators import user_passes_test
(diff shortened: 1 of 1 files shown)


biospecdb/urls.py
```diff
@@ -65,6 +66,10 @@
     path('admin/', admin.site.urls)
 ]
 
-if settings.DEBUG:
-    urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
```
jamienoss, line 69: Static files get served using whitenoise, so this was no longer needed anyhow (since #255).

PR #354, opened 2024-05-14
Build docs in apprunner build, not post-build stage

jamienoss:


jamienoss:
This worked. Items created in ``post-build`` are not persisted to ``run``. I guess this could be expected, but no-where is this stated in the [app runner docs](https://docs.aws.amazon.com/apprunner/latest/dg/config-file-ref.html) (that I could find).

PR #361, opened 2024-07-11
Temporarily limit matplotlib <3.9.0

jamienoss:
We don't directly spec this, however, it's used by plotly.

See DeepLabCut/DeepLabCut#2581 in relation to attribute error.


jamienoss:
This should resolve the issue with #360 and the [failing CI](https://github.com/RISPaDD/biospecdb/actions/runs/9882117526/job/27294352043?pr=360#step:6:106).

